PR #19, opened 2023-09-15
Improve rys roots accuracy

sunqm:


gpu4pyscf/lib/cupy_helper.py
```diff
@@ -339,250 +340,6 @@ def transpose_sum(a):
         raise RuntimeError('failed in transpose_sum kernel')
     return a
 
-LMAX_ON_GPU = 8
```
wxj6000, line 342: The following functions are deleted?

sunqm: They are duplication of code in the same file

wxj6000 (approved):

PR #20, opened 2023-09-17
added dockerfile

wxj6000:
- Added dockerfile for compilation and runtime. 
- Removed cutensor in CMakeLists.txt, it is unnecessary
- Stored dispersion correction in mf object.

### DIFF
--- a/gpu4pyscf/lib/cupy_helper/CMakeLists.txt
+++ b/gpu4pyscf/lib/cupy_helper/CMakeLists.txt
@@ -33,38 +33,4 @@ set_target_properties(cupy_helper PROPERTIES
   LIBRARY_OUTPUT_DIRECTORY ${PROJECT_SOURCE_DIR})
 find_package(CUDAToolkit)
 
-# based on https://github.com/NVIDIA/CUDALibrarySamples/blob/master/cuTENSOR/CMakeLists.txt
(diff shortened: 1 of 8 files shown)


gpu4pyscf/lib/cupy_helper/CMakeLists.txt
```diff
@@ -33,38 +33,4 @@ set_target_properties(cupy_helper PROPERTIES
   LIBRARY_OUTPUT_DIRECTORY ${PROJECT_SOURCE_DIR})
 find_package(CUDAToolkit)
 
-# based on https://github.com/NVIDIA/CUDALibrarySamples/blob/master/cuTENSOR/CMakeLists.txt
```
sunqm, line 36: is cutensor not needed?

wxj6000: cutensor is still needed as a library. But the line 'find_package(CUTENSOR)' in CMakeLists.txt is unnecessary.

sunqm: Your PR removes these code

wxj6000: Right, they are unnecessary either based on my test.

PR #27, opened 2023-09-28
Merge gradient

Walter-Feng:
This adds gradient of get_jk as well as kernels that perform contraction of gradient ERI with two density matrices. A couple changes in cmake are also made to fit into nvidia's HPC toolkits.


gpu4pyscf/df/int3c2e.py
```diff
@@ -1084,7 +1084,8 @@ def get_dh1e(mol, dm0):
     charges = mol.atom_charges()
     fakemol = gto.fakemol_for_charges(coords)
     intopt = VHFOpt(mol, fakemol, 'int2e')
-    intopt.build(1e-14, diag_block_with_triu=True, aosym=False, group_size=BLKSIZE, group_size_aux=BLKSIZE)
+    intopt.build(1e-14, diag_block_with_triu=True, aosym=False, group_size=256, group_size_aux=256)
```
wxj6000, line 1087: Please use the global variable 'BLKSIZE', avoid the hardcode constant.

gpu4pyscf/lib/cupy_helper/CMakeLists.txt
```diff
@@ -31,6 +31,49 @@ add_library(cupy_helper SHARED
 
 set_target_properties(cupy_helper PROPERTIES
   LIBRARY_OUTPUT_DIRECTORY ${PROJECT_SOURCE_DIR})
-find_package(CUDAToolkit)
+#find_package(CUDAToolkit)
```
wxj6000, line 34: It is unnecessary to find CUDAToolkit

gpu4pyscf/df/int3c2e.py
Walter-Feng: So is this line working now? If I am to adjust the size for HPC, how can I do to prevent exceeding memory usage? Or is it done automatically?

gpu4pyscf/lib/cupy_helper/CMakeLists.txt
Walter-Feng: So that's why I commented it out - is it safe to delete?

Walter-Feng: Um wait, I tried on my local machine, CUDAToolkit needs to be found to correctly link the cublas and cusolver. But on HPC it is different from CUDAToolkit so there is the if...else... command below this. I am moving this to the else part, i.e. if NVHPC cmake file is not found.

gpu4pyscf/lib/gvhf/g2e_ip1_root3.cu
```diff
@@ -0,0 +1,15905 @@
+
```
wxj6000, line 1: header here

gpu4pyscf/lib/gvhf/get_veff_driver_ip1.cu
```diff
@@ -0,0 +1,226 @@
+#include <stdio.h>
```
wxj6000, line 1: header here

gpu4pyscf/lib/gvhf/nr_jk_driver_ip1.cu
```diff
@@ -0,0 +1,248 @@
+#include <stdio.h>
```
wxj6000, line 1: header here

wxj6000 (commented):
Could you please also add an example and test cases for gradient calculation?

gpu4pyscf/df/int3c2e.py
wxj6000: Yes, BLKSIZE is defined as 256 on the top of the file.

gpu4pyscf/lib/cupy_helper/CMakeLists.txt
wxj6000: OK. I see.  Keep the one worked for you then.

PR #51, opened 2023-10-19
Optimize hessian intermediate variables

wxj6000:


gpu4pyscf/df/df.py
```diff
@@ -208,12 +208,13 @@ def cholesky_eri_gpu(intopt, mol, auxmol, cd_low, omega=None, sr_only=False):
...
-        warnings.warn("Not enough GPU memory")
+        log.debug("Not enough GPU memory")
         # TODO: async allocate memory
         mem = cupy.cuda.alloc_pinned_memory(naux * npair * 8)
-        cderi = np.ndarray([naux, npair], dtype=np.float64, order='C', buffer=mem)
-
+        try:
+            cderi = np.ndarray([naux, npair], dtype=np.float64, order='C', buffer=mem)
+        except Exception:
+            raise RuntimeError('out of CPU memory')
```
sunqm, line 217: Isn't the OOM error raised by `cupy.alloc_pinned_memory` ? It's better not capture the general `Exception` here, as it suppresses the error msg raised by numpy . It may be messages about the problem in shape.

sunqm (approved):

PR #86, opened 2024-01-19
smd grad hess

wxj6000:
- static class for PCM gradient and hessian, following pyscf 2.4
- add smd gradient and hessian
- reorg int3c2e_jk_ip1 kernel
- add benchmark for smd


gpu4pyscf/solvent/tests/test_smd.py
wxj6000, line 1: add ng = 590

PR #87, opened 2024-01-24
df uks

puzhichen:
Support UHF and UKS with density fitting.


gpu4pyscf/df/tests/test_df_uscf.py
wxj6000, line 1: separate RHF and RKS tests into two files, namely test_df_rhf.py and test_df_rks.py

puzhichen: Get!

PR #106, opened 2024-02-15
Hess nuc elec

wxj6000:
- reduce GPU memory footprint for df hessian
- move hess_nuc_elec to GPU


sunqm:
Is this PR duplicated to #110 ?

wxj6000:
No. It is not duplicated. #110 is an additional work. We shall merge this one first.

PR #109, opened 2024-02-17
Add an FCI solver fci.direct_spin1

sunqm:


gpu4pyscf/fci/direct_spin1.py
```diff
@@ -0,0 +1,191 @@
...
+    else:
+        addrb, signb = _link_index_to_addrs(link_indexb, nnorb)
+
+    threads = (TILE, TILE)
+    blocks = (1, (na+TILE-1)//TILE)
+    rest_args = (na, nb, nnorb, addra, addrb, signa, signb)
+    t1 = cp.empty((nnorb, na*TILE))
+    gt1 = cp.empty((nnorb, na*TILE))
+    if 0:
+        # Pipeline the three operations. Seems no performance improvement
```
wxj6000, line 169: Is it because _gather is running in the default stream? The default stream has to run synchronously w.r.t. other streams.

sunqm: I tried that, no help. I guess the memory bandwidth is probably saturated. I was hoping to overlap the dot operation and the other two data moving functions. However, the dot operation also occupies most of the memory bandwidth, since the product is between a tall skinny matrix and a small square matrix.

PR #110, opened 2024-02-17
Skip libxc

wxj6000:
Search libxc from local directory and site-packages, then skip libxc build.


gpu4pyscf/dft/libxc.py
```diff
@@ -18,12 +18,28 @@
...
+    try:
+        _libxc = np.ctypeslib.load_library(
+            'libxc', os.path.abspath(os.path.join(path, 'gpu4pyscf', 'lib', 'deps', 'lib')))
+        print(f'find libxc in {path}')
+    except Exception:
+        _libxc = None
+    if _libxc is not None:
+        break
+
+if _libxc is None:
```
sunqm, line 39: Shall we remove `import libxc` in `dft/__init__.py`  . Then the other module can work without installing or compiling libxc.

sunqm (approved):

gpu4pyscf/dft/libxc.py
wxj6000: OK. Let me try if it breaks anything.

wxj6000: `libxc` is imported by `numint`. We have to remove `rks` from `dft/__init__.py` for skipping `libxc`. I change the error to a warning. Other modules still can work without libxc.

PR #111, opened 2024-02-19
dispersion correction for RHF and UHF

wxj6000:


gpu4pyscf/dft/rks.py
```diff
@@ -217,6 +217,12 @@ def energy_elec(ks, dm=None, h1e=None, vhf=None):
     e1 = cupy.einsum('ij,ji->', h1e, dm).real
     ecoul = vhf.ecoul.real
     exc = vhf.exc.real
+    if isinstance(ecoul, cupy.ndarray):
```
sunqm, line 220: `.get()` method returns a 0-dimensional array. It's better to change these values to scalar variables, using `float(ecoul)` or `ecoul.get()[()]`

gpu4pyscf/scf/hf.py
```diff
@@ -370,6 +370,8 @@ def energy_elec(self, dm=None, h1e=None, vhf=None):
     if vhf is None: vhf = self.get_veff(self.mol, dm)
     e1 = cupy.einsum('ij,ji->', h1e, dm).real
     e_coul = cupy.einsum('ij,ji->', vhf, dm).real * .5
+    e1 = e1.get()
```
sunqm, line 373: Consider to convert the 0-dimensional arrays to scalars, like the one in rks.py

sunqm (approved):

PR #132, opened 2024-03-29
Refactor classes

wxj6000:
- reduce the inheritance relation with PySCF
- reuse the same function defined in PySCF
- reuse pyscf `lib.StreamObject` only
- remove patch_pyscf files, disable hessian in GPU4PySCF
- add pytest coverage


wxj6000:
Fixed `to_gpu` issue and resolved the incompatibility with pyscf in master branch. @sunqm 
https://github.com/pyscf/pyscf/pull/2149

PR #157, opened 2024-05-23
reduce pair data reads in cuda kernels

wxj6000:
- `pinv` -> `solve_j2c` in DF.
- `LINEAR_DEP_THR = 1e-6` in DF
-  `DSOLVE_LINDEP = 1e-13`.
-  reduce pair data reads in CUDA kernels


gpu4pyscf/lib/cupy_helper.py
```diff
@@ -28,7 +28,7 @@
 from gpu4pyscf.lib.cusolver import eigh, cholesky  #NOQA
 
 LMAX_ON_GPU = 7
-DSOLVE_LINDEP = 1e-15
+DSOLVE_LINDEP = 1e-14
```
wxj6000, line 31: 1e-12

PR #161, opened 2024-05-27
Fix a bug in jk cuda kernel

wxj6000:
- Remove `is_skip` calls in CUDA kernels.
- Remove `log_q_ij` and `log_q_kl` in `offset`


sunqm:
Does this modification also remove the integral screening?

wxj6000:
> Does this modification also remove the integral screening?

@sunqm The integral screening at bin level is still effective. Only the integral screening in the general CUDA kernel is removed. I feel the previous experiment is not enough. Let us re-evaluate the strategy later.

PR #170, opened 2024-06-20
numpy 2.0 compatiability

wxj6000:


gpu4pyscf/cc/ccsd_incore.py
```diff
@@ -198,8 +198,8 @@ def _direct_ovvv_vvvv(mycc, t1, t2):
     T = cupy.cuda.cublas.CUBLAS_OP_T
     one = np.ones(1)
     one_ptr = one.ctypes.data
-    x2_ptr = x2.data.ptr
-    Ht2ao_ptr = Ht2ao.data.ptr
+    x2_ptr = np.int64(x2.data.ptr)
```
sunqm, line 201: why these data need to be explicitly cast to `np.int64`?

wxj6000: Numpy 1.x used to automatically change the data type to fit the result. It will raise error in Numpy 2.0. In this case, the data type of the pointer is int32. If other integers in int32 are added, the result may not fit in int32. 

https://numpy.org/devdocs/numpy_2_0_migration_guide.html#changes-to-numpy-data-type-promotion

https://numpy.org/neps/nep-0050-scalar-promotion.html#table-comparing-new-and-old-behaviour

sunqm: I see. It's too bad that numpy adopts this downward casting.

PR #191, opened 2024-07-31
fixed a bug in rks.reset

wxj6000:
The geometry optimization of DFT with direct SCF scheme does not perform as expected due to the bug in reset function.


amartyabose:
@wxj6000 Thank you for the bug fix.  Now the geometry optimization is working.  There is a small other thing that I noticed.  At the end of the optimization run, I am getting the following error:
```
AttributeError: 'NoneType' object has no attribute 'xc_func_end'
```

This is traced to the dft/libxc.py file.  In the destructor of the XCfun class, around line 102, there are calls to xc_func_end and xc_func_free.  These need to be enclosed inside a check for _libxc being not null:

```python
    def __del__(self):
        if self.xc_func is None:
            return
        if _libxc is not None:
            _libxc.xc_func_end(self.xc_func)
            _libxc.xc_func_free(self.xc_func)
```

Would you change this as well?

wxj6000:
@amartyabose Interesting. `_libxc` is the loaded library for xc potentials.  It should not be `None` as the geometry optimization is already working. The error does not appear on my side. It does not appear in the tests either. Can you please provide some info about your environment?

amartyabose:
@wxj6000 It is the same input as the one I had provided for the geometry minimization issue.  This issue about `_libxc` being `None` is only arising when I am calling the geometry minimization routine.  For a regular scf single point calculation it is cleaning up just fine.

```python
from datetime import datetime
import numpy as np

from pyscf import gto
from pyscf.geomopt.geometric_solver import optimize
from gpu4pyscf import dft

import warnings
warnings.filterwarnings("ignore")

mol = '''C                 -1.43530661   -0.38248086    0.00007236;
 C                 -0.03390667   -0.38255097    0.00050137;
 C                  0.66685425    0.83106172   -0.00006381;
 C                 -0.03378475    2.04474452   -0.00105877;
 C                 -1.43518468    2.04481464   -0.00148574;
 C                 -2.13594561    0.83120195   -0.00092082;
 H                 -1.97035312   -1.30910108    0.00050287;
 H                  0.50104676   -1.30922473    0.00125995;
 H                  1.73685420    0.83100818    0.00026331;
 H                  0.50126177    2.97136475   -0.00149182;
 H                 -1.97013811    2.97148840   -0.00224291;
 H                 -3.20594556    0.83125549   -0.00124699
'''

mol_benzene = gto.M(
    atom = mol,
    basis = '6-31g'
)

rks_benzene = dft.RKS(mol_benzene)
rks_benzene.xc = 'wb97x_d'
rks_benzene.grids.level = 8

start_time = datetime.now()
rks_benzene.kernel()
print(f'Total time taken for single point energy = {(datetime.now() - start_time).total_seconds()} seconds')

start_time = datetime.now()
opt_benzene = optimize(rks_benzene)
print(f'Total time taken for geometry optimization = {(datetime.now() - start_time).total_seconds()} seconds')

print(opt_benzene.tostring())

rks_benzene = dft.RKS(opt_benzene)
rks_benzene.xc = 'wb97x_d'
rks_benzene.grids.level = 8

start_time = datetime.now()
rks_benzene.kernel()
print(f'Total time taken for single point energy = {(datetime.now() - start_time).total_seconds()} seconds')
```

wxj6000:
@amartyabose I am still not able to reproduce the issue. But I follow the your suggestion, adding the check in `__del__`. Please let me know if you find new evidence.

PR #203, opened 2024-08-29
Add chkfile support for pysisyphus

henryw7:
Connect the chkfile related functions from pyscf into gpu4pyscf.

### DIFF
--- a/gpu4pyscf/scf/hf.py
+++ b/gpu4pyscf/scf/hf.py
@@ -17,6 +17,8 @@
 
 import copy
 import ctypes
+import sys
+import h5py
 import numpy as np
 import cupy
 import scipy.linalg
@@ -450,7 +452,7 @@ def _kernel(mf, conv_tol=1e-10, conv_tol_grad=None,
             local_variables = locals()
             for key in local_variables:
                 value = local_variables[key]
-                if (type(value) == cupy.ndarray):
+                if (type(value) is cupy.ndarray):
                     local_variables[key] = cupy.asnumpy(value)
             mf.dump_chk(local_variables)
 
--- a/gpu4pyscf/scf/hf.py
+++ b/gpu4pyscf/scf/hf.py
@@ -423,6 +423,8 @@ def _kernel(mf, conv_tol=1e-10, conv_tol_grad=None,
         # Explicit overwrite the mol object in chkfile
         # Note in pbc.scf, mf.mol == mf.cell, cell is saved under key "mol"
         chkfile.save_mol(mol, mf.chkfile)
+    print(mf.chkfile)
+    exit()
     
     for cycle in range(mf.max_cycle):
         t0 = log.init_timer()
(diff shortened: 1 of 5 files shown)


gpu4pyscf/scf/hf.py
```diff
@@ -419,6 +418,11 @@ def _kernel(mf, conv_tol=1e-10, conv_tol_grad=None,
         _, mf_diis.Corth = mf.eig(fock, s1e)
     else:
         mf_diis = None
+
+    if dump_chk and mf.chkfile:
```
sunqm, line 422: `dump_chk` can be disabled by default. dumping chkfile can noticeably slow down the scf function.

henryw7: I think it's a better idea to set `self.chkfile = None` at __init__ function of SCF class (which requires a copy of __init__ function from pyscf.hf.SCF to gpu4pyscf.hf.SCF). If we change the function signature of this internal kernel function, it'll be very hard to turn in on from pysisyphus side.

sunqm: All work. This flag just needs to be muted by default.

gpu4pyscf/scf/tests/test_rhf.py
```diff
@@ -264,11 +265,23 @@ def test_rhf_d4(self):
...
+        mf = scf.RHF(mol)
+        mf.chkfile = ftmp.name
+        mf.kernel()
+        dm_stored = mf.make_rdm1(mf.mo_coeff, mf.mo_occ)
+        dm_stored = cupy.asnumpy(dm_stored)
+
+        mf_copy = scf.RHF(mol)
+        mf_copy.chkfile = ftmp.name
+        dm_loaded = mf_copy.init_guess_by_chkfile()
+        assert np.allclose(dm_stored, dm_loaded, atol = 1e-14)
```
wxj6000, line 279: 1e-14 is probably too strict. 1e-8 or 1e-9 should be enough for this test.

gpu4pyscf/scf/tests/test_uhf.py
```diff
@@ -272,13 +273,28 @@ def test_uhf_d4(self):
...
+        mf = scf.UHF(mol)
+        mf.chkfile = ftmp.name
+        mf.kernel()
+        dma_stored, dmb_stored = mf.make_rdm1(mf.mo_coeff, mf.mo_occ)
+        dma_stored, dmb_stored = cupy.asnumpy(dma_stored), cupy.asnumpy(dmb_stored)
+
+        mf_copy = scf.UHF(mol)
+        mf_copy.chkfile = ftmp.name
+        dma_loaded, dmb_loaded = mf_copy.init_guess_by_chkfile()
+        assert np.allclose(dma_stored, dma_loaded, atol = 1e-14)
```
wxj6000, line 288: tol for density matrix can be 1e-5 or 1e-6

gpu4pyscf/scf/tests/test_rhf.py
henryw7: I added a comment about it. The two density matrices are not from two SCF calculations, but from the MO coefficients before storage and after storage and reload. This threshold suppose to be zero, I just allow some numerical noise.

wxj6000 (approved):

PR #208, opened 2024-09-14
Add esp charge and resp charge

wxj6000:


gpu4pyscf/pop/esp.py
```diff
@@ -0,0 +1,222 @@
...
+    1.36, # Mg
+    1.24, # Al,
+    1.17, # Si,
+    1.80, # P,
+    1.75, # S,
+    1.70]) # Cl
+
+def unit_surface(n):
+    '''
+    Generate spherical harmonics grid points on unit sphere
```
sunqm, line 55: 1. It's not spherical harmonics grids. Just one kind of "uniform" grids. Is it right? 
Is there a link or a reference for the grids generation scheme.

2. Why not using Lebedev grids?

gpu4pyscf/pop/esp.py
```diff
@@ -0,0 +1,222 @@
...
+
+        ux.append(x)
+        uy.append(y)
+        uz.append(z*np.ones_like(x))
+    
+    ux = np.concatenate(ux)
+    uy = np.concatenate(uy)
+    uz = np.concatenate(uz)
+
+    return np.array([ux[:n], uy[:n], uz[:n]]).T
```
sunqm, line 84: This grid scheme does not necessary produce n grids, right? For example, if n = 30, the output would produce only 23 grids.

gpu4pyscf/pop/esp.py
wxj6000: 1. Right. The reference is on the top of the file.
2. For reproducibility. The 'uniform' was used qchem, psi4, and GAMESS. Q-Chem also provides the support of Lebedev grids. We can add the support if requested.

gpu4pyscf/pop/esp.py
wxj6000: Right. It generally produces less than n points. Let me add some comments here.

PR #224, opened 2024-10-12
Optimize nuclear gradients using new Rys kernel

sunqm:
* Optimize the ERI kernel in gradients computation
* Add grid_response for gradients methods.


gpu4pyscf/grad/rhf.py
```diff
@@ -15,216 +15,147 @@
...
+        vhfopt = _VHFOpt(mol).build()
+
+    mol = vhfopt.mol
+    nao, nao_orig = vhfopt.coeff.shape
+
+    dm = cp.asarray(dm, order='C')
+    dms = dm.reshape(-1,nao_orig,nao_orig)
     n_dm = dms.shape[0]
-    scripts = []
+    dms = cp.einsum('pi,nij,qj->npq', vhfopt.coeff, dms, vhfopt.coeff)
```
wxj6000, line 58: sandwich_dot?

gpu4pyscf/grad/rks.py
```diff
@@ -318,15 +328,16 @@ def get_vxc_full_response(ni, mol, grids, xc_code, dms, relativity=0, hermi=1,
...
+    natm = mol.natm
     mol = None
     _sorted_mol = opt._sorted_mol
     coeff = cupy.asarray(opt.coeff)
     nao, nao0 = coeff.shape
     dms = cupy.asarray(dms)
-    dms = [cupy.einsum('pi,ij,qj->pq', coeff, dm, coeff)
-           for dm in dms.reshape(-1,nao0,nao0)]
+    assert dms.ndim == 2
+    dms = cupy.einsum('pi,ij,qj->pq', coeff, dms, coeff)
```
wxj6000, line 338: sandwich_dot

gpu4pyscf/grad/tests/test_uks_grad.py
```diff
@@ -26,40 +26,35 @@
...
+                       output='/dev/null', verbose=1)
 
 def tearDownModule():
     global mol_sph, mol_cart
     mol_sph.stdout.close()
     mol_cart.stdout.close()
     del mol_sph, mol_cart
 
+# FIXME: Why is the difference between CPU and GPU so large?
+# tol=1e-6 is not acceptable
```
wxj6000, line 47: The consistency between CPU and GPU is at the level of 1e-10. We should be able to set it to 1e-8.

wxj6000 (approved):

gpu4pyscf/grad/tests/test_uks_grad.py
sunqm: Some tests show an error ~1e-6 . I don't think this is caused by bugs. I'm not clear the reason of the difference. They should match to a very high accuracy for such small systems. I will leave this comment there and further research on this problem.

PR #234, opened 2024-11-02
Calculate `dm` if it is not given in `solvent_mf.get_fock()`

wxj6000:


wxj6000:
Same issue in https://github.com/pyscf/pyscf/pull/2492

sunqm (approved):

PR #239, opened 2024-11-08
Optimize DF-TDA by using the MO-based DF.get_jk implementation

sunqm:


gpu4pyscf/tdscf/tests/test_tduks.py
```diff
@@ -171,7 +171,7 @@ def test_tda_m06l(self):
         es = td.kernel(nstates=5)[0]
         ref = td.to_cpu().kernel(nstates=5)[0]
         self.assertAlmostEqual(abs(es - ref[:5]).max(), 0, 8)
-        self.assertAlmostEqual(lib.fp(es), -0.7531765338077471, 6)
+        self.assertAlmostEqual(lib.fp(es), -0.7530329968766932, 6)
```
wxj6000, line 174: Why are those values changed?

sunqm: Because the test was created using direct scf ground state, computed without density fitting. They are changed to density fitting here.

wxj6000 (approved):

PR #246, opened 2024-11-13
Initialize PBC-HF modules

sunqm:
Added
* The gamma point PBC-HF base classes.
* Gamma point PBC-RKS base classes
* The FFTDF class for GPW integral method.


gpu4pyscf/pbc/__init__.py
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env python
+#
+# Copyright 2024 The PySCF Developers. All Rights Reserved.
```
wxj6000, line 3: GPU4PySCF Developers

wxj6000 (approved):

PR #262, opened 2024-12-02
Minor optimization in Rys integrals

wxj6000:
- Reuse rys roots in gvhf-rys.
- Reduce local memory in `int3c2_ipip1` and `int3c2e_ipip2`.


gpu4pyscf/lib/gvhf-rys/rys_roots.cuh
```diff
@@ -416,11 +416,6 @@ __device__ static double ROOT_LARGEX_W_DATA[] = {
 5.2158612689910977e-01,
 };
```
sunqm, line 418: Please keep these size definitions

wxj6000: These variables are defined at the beginning of the file. They are redefined here.

sunqm (approved):

PR #263, opened 2024-12-02
Clean up one electron integral python code according to new convention

henryw7:


wxj6000 (approved):

gpu4pyscf/gto/moleintor.py
```diff
@@ -265,21 +245,18 @@ def get_int3c1e(mol, grids, direct_scf_tol):
...
+        ao_idx = np.argsort(intopt._ao_idx)
         grid_idx = np.arange(ngrids_of_split)
         int3c_grid_slice = int3c_grid_slice[np.ix_(grid_idx, ao_idx, ao_idx)]
 
         int3c_grid_slice.get(out = int3c[i_grid_split : i_grid_split + ngrids_of_split, :, :])
 
     return int3c
 
-def get_int3c1e_charge_contracted(mol, grids, charges, direct_scf_tol):
+def get_int3c1e_charge_contracted(mol, grids, charges, intopt):
```
wxj6000, line 256: Set intopt as optional argument.

gpu4pyscf/gto/moleintor.py
```diff
@@ -346,26 +323,26 @@ def get_int3c1e_charge_contracted(mol, grids, charges, direct_scf_tol):
...
-    return cp.asnumpy(int1e)
+    return int1e
 
-def get_int3c1e_density_contracted(mol, grids, dm, direct_scf_tol):
+def get_int3c1e_density_contracted(mol, grids, dm, intopt):
     omega = mol.omega
     assert omega >= 0.0, "Short-range one electron integrals with GPU acceleration is not implemented."
 
     dm = cp.asarray(dm)
+    if dm.ndim == 3:
```
wxj6000, line 336: dm = dm[0] + dm[1] for unrestricted case

gpu4pyscf/gto/moleintor.py
henryw7: This is an internal function, the interface function https://github.com/pyscf/gpu4pyscf/blob/03d3044e8abb53bd8eac2374fcb91231a0d03b96/gpu4pyscf/gto/moleintor.py#L415
has intopt as optional. Should I use underscore prefix for this function?

gpu4pyscf/gto/moleintor.py
henryw7: `cp.einsum("ijk->jk", dm)` is doing `dm[0] + dm[1]` logic, what's the concern?

gpu4pyscf/gto/moleintor.py
wxj6000: OK. This should be enough.

gpu4pyscf/gto/moleintor.py
wxj6000: It is a bit too general. But it is OK.

PR #284, opened 2024-12-12
Remove int3c2e kl operation, which causes an edge case bug

henryw7:
I think the problem is the default gaussian charge exponent for point charge is too big and caused numerical problem.
`def fakemol_for_charges(coords, expnt=1e16)`

### DIFF
--- a/gpu4pyscf/df/tests/test_int3c2e.py
+++ b/gpu4pyscf/df/tests/test_int3c2e.py
@@ -14,10 +14,11 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import pyscf
-from pyscf import df
+from pyscf import df, gto
 from pyscf.gto.moleintor import getints, make_cintopt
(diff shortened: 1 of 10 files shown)


henryw7:
Nice it is reproduced also on the test machine, the behavior is the same as on V100 and T4, only one element is different by a huge amount:
```Mismatched elements: 1 / 6 (16.7%)
Max absolute difference among violations: 2.99363254
Max relative difference among violations: 1.02413183
 ACTUAL: array([[-0.062726,  0.07054 ],
       [-0.097968, -0.076773],
       [-0.098293, -0.077027]])
 DESIRED: array([[-0.062726, -2.923093],
       [-0.097968, -0.076773],
       [-0.098293, -0.077027]])
```

wxj6000:
Nice catch! You can move this to issues. We probably need to add more assertions globally to avoid any underflows.

sunqm (approved):

PR #287, opened 2024-12-17
Replace min(blksize, MIN_BLKSIZE) by max(blksize, MIN_BLKSIZE)

DerkKooi:
`numint.py` contains the line
```python
blksize = min(blksize, MIN_BLK_SIZE)
```

Now either `min` or `MIN` here should be replaced by `max`/`MAX` to make this correct. My suggestion is that it should be `blksize = max(blksize, MIN_BLK_SIZE)` as right now the behavior is such that that much too small block sizes are selected for small systems. On the other hand, a minimum block size seems like it would run out of memory for larger systems.

### DIFF
--- a/gpu4pyscf/dft/numint.py
+++ b/gpu4pyscf/dft/numint.py
@@ -91,7 +91,7 @@ def eval_ao(mol, coords, deriv=0, shls_slice=None, nao_slice=None, ao_loc_slice=
     coords = cupy.asarray(coords.T, order='C')
     comp = (deriv+1)*(deriv+2)*(deriv+3)//6
     stream = cupy.cuda.get_current_stream()
-    
+
     # ao must be set to zero due to implementation
     if deriv > 1:
         if out is None:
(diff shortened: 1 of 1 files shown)


DerkKooi:
CC: @awvwgk @tvogels @jhrmnn

wxj6000:
Hi @DerkKooi, thank you for opening the discussion.

When your device has less than 16 GB memory, the default `MIN_BLKSIZE` is 4096 which is pretty small. We probably can use  `MIN_BLKSIZE = 16384` for better performance. But the configurations of those devices can be very diverse, `MIN_BLKSIZE = 4096` is just for safety.

`MIN_BLKSIZE` is for the sparsity of AO on grids. Smaller `blksize` benefits the sparsity of atomic orbitals on grids, therefore improves the performance. 
On the other hand, the values of atomic orbitals on grids should fit in the memory. Based on these two constraints, we set `blksize = min(blksize, MIN_BLK_SIZE)`. 

Can you provide your device info? And how the block sizes are selected for a small system?

PR #297, opened 2024-12-31
PBC 3c2e short-range ERIs

sunqm:


examples/40-all_electron_scf.py
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env python
```
wxj6000, line 1: License header

gpu4pyscf/pbc/gto/cell.py
```diff
@@ -0,0 +1,35 @@
+import numpy as np
```
wxj6000, line 1: License header

wxj6000 (approved):

PR #307, opened 2025-01-17
map and reduce functions for multi-gpu runtime

sunqm:


gpu4pyscf/lib/logger.py
```diff
@@ -29,66 +26,63 @@
...
 def init_timer(rec):
-    if rec.verbose >= TIMER_LEVEL:
-        e0 = cupy.cuda.Event()
-        e0.record()
-        return (process_clock(), perf_counter(), e0)
-    elif rec.verbose >= DEBUG:
-        return (process_clock(), perf_counter())
-    else:
-        return process_clock(),
+    e0 = cupy.cuda.Event()
```
wxj6000, line 34: Removing this will introduce the overhead of creating an event and calling recording. We will need to check if init_timer is abused.

wxj6000 (approved):

PR #318, opened 2025-01-26
Add 3c driver

wxj6000:


jeanwsr:
B97-3c and r2scan-3c should use a jfit auxbasis. Looks like the current default is a jkfit auxbasis?

wxj6000:
> B97-3c and r2scan-3c should use a jfit auxbasis. Looks like the current default is a jkfit auxbasis?

Nice catch! B97-3c and r2scan-3c are pure DFT. We should use jfit auxbasis. It is better to add it to the config file. The PR is still in progress, and probably will not be merged. I open it for discussion and exploring the support of 3c methods in PySCF ecosystem.

wxj6000:
@sunqm Ready for review.

sunqm (approved):

PR #331, opened 2025-02-10
Record a trap in current dft 3c driver code

henryw7:


wxj6000 (approved):

Tillsten:
Short question, should a similar change also be applied to the 3c optimatizion driver?

Issue #343, opened 2025-02-28
Deprecate the support of SM60, SM61 and SM62 architectures in PyPi packages


wxj6000:
The following GPU cards will not be supported in the future (tentatively GPU4PySCF v1.4).
- Quadro GP100, Tesla P100
- GTX 1080, GTX 1070, GTX 1060, and so on.

PR #344, opened 2025-03-03
Several maintenance issues

wxj6000:
- Deprecate SM60 (https://github.com/pyscf/gpu4pyscf/issues/343)
- Dump grids info and xc info (https://github.com/pyscf/gpu4pyscf/issues/341)
- Split the benchmark workflow into three jobs
- Pass sqrt(lindep) instead of lindep in tddft eigh solver, which caused overflow/underflow issues


gpu4pyscf/tdscf/_lr_eig.py
```diff
@@ -824,7 +824,7 @@ def _symmetric_orth_real(xt, lindep=1e-6):
     # s = [[s11, s21.conj().T],
     #      [s21, s11.conj()  ]]
     e, c = np.linalg.eigh(s11)
-    mask = e > lindep**2
+    mask = e > lindep_sqrt
```
sunqm: missing **2

PR #352, opened 2025-03-07
Reduce GPU memory footprint in SCF

sunqm:
* [x] Promptly release the attributes of the tagged arrays (dm, veff)
* [x] Promptly release the unused variables within the SCF iteration
* [x] Only save the lower triangular part of the matrices in DIIS
* [x] Hold certain transformation matrices on host memory
   - [x] vhfopt.coeff
   - [x] diis.Corth
* A separated direct SCF driver for large-size systems:
   - [x] overlap matrix computed on the fly
   - [x] Save only the lower triangular part of hcore, veff matrices
* Bugfix: Use orthogonal bases in DIIS error vector construction.

### DIFF
--- a/gpu4pyscf/dft/__init__.py
+++ b/gpu4pyscf/dft/__init__.py
@@ -11,7 +11,7 @@ def KS(mol, xc='LDA,VWN'):
     else:
         return UKS(mol, xc)
 
-def RKS(mol, xc):
+def RKS(mol, xc='LDA,VWN'):
     from gpu4pyscf.lib.cupy_helper import get_avail_mem
     from . import rks_lowmem
     mem = get_avail_mem()
--- a/gpu4pyscf/dft/__init__.py
+++ b/gpu4pyscf/dft/__init__.py
@@ -1,5 +1,6 @@
 from . import rks
-from .rks import KohnShamDFT
+from .rks import RKS, KohnShamDFT
+from .rks_lowmem import RKS as LRKS
 from .uks import UKS
 from .gks import GKS
 from .roks import ROKS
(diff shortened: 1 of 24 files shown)


gpu4pyscf/dft/rks_lowmem.py
wxj6000, line 1: Is it necessary to maintain two classes for low memory object and high memory object? Overall, we should reduce memory footprint for all the cases at a reasonable cost (labor cost and computational cost). I feel that a proper design of caching mechanism probably is enough.

sunqm: The two kernels are not compatible. The low-mem version breaks certain APIs comparing to the normal version. This class is not suitable for other methods to inherit. Merging the two kernels into one will complicate the normal kernel. More data type checks or shape checks and if-else branches have to be added into the normal kernel. This will make the normal kernel very messy.

sunqm: In the low-mem kernel, some matrices are assumed to be real, symmetric or a squared matrix. This is only true for molecule calculations.

gpu4pyscf/lib/cusolver.py
```diff
@@ -127,18 +127,19 @@ def eigh(h, s):
             fn = libcusolver.cusolverDnZhegvd_bufferSize
         status = fn(
             _handle,
```
wxj6000, line 129: The argtypes of these cusolver functions are casted already. No need to cast it again.

gpu4pyscf/dft/__init__.py
```diff
@@ -10,3 +10,13 @@ def KS(mol, xc='LDA,VWN'):
         return RKS(mol, xc)
     else:
         return UKS(mol, xc)
+
+def RKS(mol, xc='LDA,VWN'):
```
wxj6000, line 14: Should we explicitly define a `LRKS` class to distinguish low memory RKS and regular RKS? User will need to explicitly import LRKS for the low memory functionality, instead of automatic switch?

wxj6000 (approved):

PR #356, opened 2025-03-11
Remove coeff storage

henryw7:


sunqm:
Since the transformation matrix is highly sparse, does it make sense to store only the non-zero elements and their coordinates, then constructing the dense coeff matrix as needed? Maybe this is not as efficient as the sparse transformation you implemented. However, this solution is much simpler to understand.

wxj6000:
> Since the transformation matrix is highly sparse, does it make sense to store only the non-zero elements and their coordinates, then constructing the dense coeff matrix as needed? Maybe this is not as efficient as the sparse transformation you implemented. However, this solution is much simpler to understand.

As long as we name the functions clearly and document the functionality, it should be understandable. But constructing coeff matrix on-the-fly probably works more effectively. We can cache coeff matrix when the system is small.

gpu4pyscf/scf/jk.py
```diff
@@ -176,6 +184,165 @@ def build(self, group_size=None, verbose=None):
...
+
+        l_ctr_count = np.asarray(self.l_ctr_offsets[1:] - self.l_ctr_offsets[:-1], dtype = np.int32)
+        l_ctr_l = np.asarray(self.uniq_l_ctr[:,0].copy(), dtype = np.int32)
+        self.l_ctr_pad_counts = np.asarray(self.l_ctr_pad_counts, dtype = np.int32)
+        cupy_ao_idx = self.cupy_ao_idx
+        stream = cp.cuda.get_current_stream()
+
+        # ref = cp.einsum("ij,qjk,kl->qil", self.coeff, spherical_matrix, self.coeff.T)
+
+        out = cp.empty((counts, n_cartesian, n_cartesian), order = "C")
```
henryw7, line 246: This out variable suppose to be filled with zero first, to keep the pad space zero.

wxj6000: The padded zeros are supposed to be used. It can be any values, right?

wxj6000 (approved):

PR #362, opened 2025-03-26
Reduce C-PCM energy and gradient memory requirement

henryw7:
Now C-PCM only store one N_grid^2 matrix (K_LU) permanently in PCM energy and gradient calculation. At C-PCM construction stage it requires another N_grid^2 size of workspace for LU. At gradient calculation step, another N_grid * 6 * N_atom of space (dF, dSii) will be required temporarily.

### DIFF
--- a/gpu4pyscf/solvent/hessian/pcm.py
+++ b/gpu4pyscf/solvent/hessian/pcm.py
@@ -333,14 +333,16 @@ def analytical_hess_qv(pcmobj, dm, verbose=None):
     t1 = log.timer_debug1('solvent hessian d(dI/dx * q)/dx contribution', *t1)
     return d2e
 
-def einsum_ij_Adj_Adi_inverseK(K, Adj_term):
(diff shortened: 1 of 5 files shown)


gpu4pyscf/solvent/hessian/pcm.py
```diff
@@ -333,14 +333,16 @@ def analytical_hess_qv(pcmobj, dm, verbose=None):
     t1 = log.timer_debug1('solvent hessian d(dI/dx * q)/dx contribution', *t1)
     return d2e
 
-def einsum_ij_Adj_Adi_inverseK(K, Adj_term):
+def einsum_ij_Adj_Adi_inverseK(pcmobj, Adj_term, K_transpose = False):
```
wxj6000, line 336: This function and next function are quite similar. Consider to define one function which support batched solve.

sunqm (approved):

wxj6000 (approved):

PR #391, opened 2025-04-23
vv10 analytical hessian

henryw7:
Grid response is implemented for Fock matrix first derivative, but not for energy second derivative.


gpu4pyscf/hessian/rks.py
```diff
@@ -575,20 +577,471 @@ def get_nlc_de(grad_obj, dm):
...
+    d2mu_dr2[2,1,:,:] = ao[YZ, :, :]
+    d2mu_dr2[2,2,:,:] = ao[ZZ, :, :]
+    return d2mu_dr2
+
+def get_d3mu_dr3(ao):
+    assert ao.ndim == 3
+    nao = ao.shape[1]
+    ngrids = ao.shape[2]
+
+    d3mu_dr3 = cupy.empty([3, 3, 3, nao, ngrids])
```
wxj6000, line 617: Huge memory consumption is here. How large can we calculate roughly?

wxj6000 (approved):

PR #396, opened 2025-05-01
vv10 polarizability

henryw7:


gpu4pyscf/scf/_response_functions.py
```diff
@@ -16,6 +16,7 @@
 from pyscf import lib
 from gpu4pyscf.lib import logger
 from gpu4pyscf.scf import hf, uhf, rohf
+from gpu4pyscf.hessian.rks import nr_rks_fnlc_mo
 
 def _gen_rhf_response(mf, mo_coeff=None, mo_occ=None,
                       singlet=None, hermi=0, grids=None, max_memory=None):
```
sunqm, line 22: Add an option "with_nlc" to control whether to apply nlc. If implementation of mf.do_nlc is not available in other response functions (UHF, GHF), this option can be false by default

gpu4pyscf/properties/polarizability.py
sunqm, line 1: Update `vresp = mf.gen_response(with_nlc=True)` . Don't forget to update the signature of the _response_function for UHF, GHF

gpu4pyscf/scf/_response_functions.py
henryw7: This function has checked if the mf is a HF or KS object:
https://github.com/pyscf/gpu4pyscf/blob/03ee213d90b899a3f5ccf05b5bb3eca56da05f5f/gpu4pyscf/scf/_response_functions.py#L34
And `do_nlc()` method is only called if mf is a KS object.
Also I've disabled the unrestricted case here:
https://github.com/bytedance/gpu4pyscf/blob/0241990ec99cb7a34c2d86684469d82fc4e567cc/gpu4pyscf/scf/_response_functions.py#L167
As well as other untested cases (I don't think GHF/GKS is supported). In conclusion, I don't think the new keyword is necessary.

sunqm: Consider the case of TDDFT, the tddft solver also calls this response function. with_nlc option allows the tddft solver to skip the expensive nlc part. In the nuclear hessian, we can also consider to disable nlc part by default.

wxj6000: I feel 'support with_nlc = False' is a new feature. It will require to design the corresponding APIs. Disabling nlc part by default is a bit misleading. @henryw7 will do some benchmarks and optimization after this PR. We can also reevaluate the necessity of `with_nlc` by then.

gpu4pyscf/scf/_response_functions.py
```diff
@@ -160,6 +172,11 @@ def vind(dm1):
             else:
                 v1 = ni.nr_uks_fxc(mol, grids, mf.xc, dm0, dm1, 0, hermi,
                                    rho0, vxc, fxc, max_memory=max_memory)
+                if mf.do_nlc():
+                    if with_nlc:
+                        raise NotImplementedError("NLC not supported")
```
sunqm, line 177: Does this work for UKS ? `v1 += nr_rks_fnlc_mo(mf, mol, None, None, dm1[0]+dm1[1], return_in_mo = False)`

sunqm (approved):

gpu4pyscf/scf/_response_functions.py
henryw7: Probably, I'll enable and test UHF case in a separate PR for TDDFT

PR #403, opened 2025-05-09
Better default for eval_polarizability() function

henryw7:


henryw7:
For most molecules in benchmarks/molecules/organic, wb97m-v/def2-svp (or def2-tzvpd) it takes around 20 steps to achieve 1e-7 threshold in CPHF. The new default is safer and doesn't introduce too much damage in accuarcy (tests still passing).

wxj6000 (approved):

PR #406, opened 2025-05-13
Add geometry optimization for TDDFT using PCM

puzhichen:


wxj6000 (approved):

gpu4pyscf/grad/tdrhf.py
```diff
@@ -261,6 +261,9 @@ def __call__(self, mol_or_geom, state=None, **kwargs):
         td_scanner(mol)
         assert td_scanner.device == 'gpu'
         assert self.device == 'gpu'
+        if getattr(self.base, 'with_solvent', None):
```
sunqm, line 264: with_solvent should not be updated this way. solvent should be updated by the `self.reset(mol)` command above

PR #409, opened 2025-05-14
fix the codes of reset PCM in TDDFT optimization

puzhichen:


sunqm:
I'll update the as_scanner method in another PR

PR #417, opened 2025-05-26
feat: TDDFT-ris store Tpq in RAM

John-zzh:
I think the current setup already makes the best of old `intopt` and `get_int3c2e_slice` engine.

- [ 1st commit ]

dos2unix operation is the only change in this commit. You can trust it and approve.

- [other commits] 

1.  Add option to store Tia Tij Tab in RAM. Milestone: 360 atoms, 6284 nbf wb97x-d3bj TDA-ris 10 states on A100, 40 sec if in GPU-mem; 100 sec if in RAM; TDDFT-ris 130 sec if in RAM. 
2.  cleanup multiple lines of function args
3.  sort eri2c_inv only once
4.  remove temporarily useless codes, such as UKS.   

what I have tested out :
1. on the fly generate Tia is ~20 times slower than pre-calc and store it in RAM
2. a naive multi-gpu implementation make it even slower

TODO:
1.  change to `int3c2e_bdiv` engine 
3.  test 1000 atoms system; still looking for interesting benchmark set 

x.  Maybe add UKS in the future, or maybe just head towards ROKS.


gpu4pyscf/tdscf/ris.py
wxj6000, line 1: Git is still not tracking the changes.

PR #419, opened 2025-05-28
feat: TDDFT-ris store Tia Tij Tab in RAM

John-zzh:
I think the current setup already makes the best of old `intopt` and `get_int3c2e_slice` engine.

- [changes ] 
- 
1.  Add option to store Tia Tij Tab in RAM. 
2. Milestone: wb97x-d3bj/def2-ZTVP TDA-ris  360 atoms, 6284 nbf  10 states on A100, 40 sec if in GPU-mem; 100 sec if in RAM; TDDFT-ris 130 sec if in RAM.    600+atoms def2-SVP is also possible, but still looking for stable SCF solution first.
3.  cleanup multiple lines of function args
4.  sort eri2c_inv and mo_coeff before AO->MO transformation, instead of sorting Puv
5.  remove temporarily useless codes, such as UKS.   

- [ what I have tested out and did not work well (those codes are not in commits) ] 
1. on the fly generate Tia is ~20 times slower than pre-calc and store it in GPU/RAM
2. a naive multi-gpu implementation make it even slower

TODO:
1.  test  new`int3c2e_bdiv` engine 
3.  test 1000 atoms system; still looking for interesting benchmark set 

x.  Maybe add UKS in the future, or maybe just head towards ROKS.

### DIFF
--- a/gpu4pyscf/tdscf/ris.py
+++ b/gpu4pyscf/tdscf/ris.py
@@ -25,6 +25,7 @@ from pyscf.data.nist import HARTREE2EV
 from gpu4pyscf.lib import logger
 
 
+
 CITATION_INFO = """
 Please cite the TDDFT-ris method:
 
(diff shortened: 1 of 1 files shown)


gpu4pyscf/tdscf/ris.py
```diff
@@ -778,38 +784,46 @@ def build(self):
...
+                self._JK_share_aux = True
 
-    def get_mdpol(self):
-        '''
-        magnatic dipole m
-        '''
-        int_rxp = self.mol.intor('int1e_cg_irxp' + self.eri_tag, comp=3, hermi=2)
-        int_rxp = cp.asarray(int_rxp, dtype=cp.float32 if self.single else cp.float64)
+            else:
+                log.info('J and K uese different aux basis, or RSH omega != 0')
```
jeanwsr, line 798: "use"?

John-zzh: ok, revised

gpu4pyscf/tdscf/ris.py
```diff
@@ -256,28 +236,36 @@ def compute_Tpq_on_gpu_general(mol, auxmol, C_p, C_q, lower_inv_eri2c,
...
+    nao = mol.nao
+    naux = auxmol.nao
 
     intopt = VHFOpt(mol, auxmol, 'int2e')
-    intopt.build(aosym=aosym, group_size=group_size, group_size_aux=group_size_aux)
+    intopt.build(aosym=True, group_size=group_size, group_size_aux=group_size_aux,verbose=mol.verbose)
 
-    nao = mol.nao
-    naux = auxmol.nao
+    C_p = C_p[intopt._ao_idx,:].copy()
```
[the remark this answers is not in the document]
John-zzh: I expected that C_p[intopt._ao_idx,:] merely return the view of C_p, which leads to non-contiguous memory layout,  C_p involves in contract operation many times later.
But maybe nao is small enough so that the all elements are always in cache no matter how they sorted?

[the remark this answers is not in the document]
sunqm: fancy indexing always returns an array in contiguous storage

gpu4pyscf/tdscf/ris.py
```diff
@@ -256,28 +231,36 @@ def compute_Tpq_on_gpu_general(mol, auxmol, C_p, C_q, lower_inv_eri2c,
...
 
-    nao = mol.nao
-    naux = auxmol.nao
+    C_p = C_p[intopt._ao_idx,:]
+    C_q = C_q[intopt._ao_idx,:]
 
     siz_p = C_p.shape[1]
     siz_q = C_q.shape[1]
 
+    upper_inv_eri2c = lower_inv_eri2c[intopt._aux_ao_idx,:][:,intopt._aux_ao_idx].T.copy()
```
sunqm, line 246: 1. `lower_inv_eri2c[intopt._aux_ao_idx[:,None],intopt._aux_ao_idx]`
2. .T.copy() is not necessary. einsum can handle either contiguous storage efficiently

gpu4pyscf/tdscf/ris.py
```diff
@@ -1627,17 +1178,12 @@ def kernel(self):
             if self.RKS:
                 TDDFT_hybrid_MVP, hdiag = self.gen_RKS_TDDFT_hybrid_MVP()
 
-            elif self.UKS:
-                TDDFT_hybrid_MVP, hdiag = self.get_UKS_TDDFT_MVP()
+            #  TODO: UKS
```
sunqm, line 1181: raise NotImplementedError

PR #422, opened 2025-06-02
Generate int3c2e in batches

sunqm:
Add functions to generate 3c2e integrals in batches along the auxiliary dimension. @John-zzh

### DIFF
--- a/gpu4pyscf/df/int3c2e_bdiv.py
+++ b/gpu4pyscf/df/int3c2e_bdiv.py
@@ -324,7 +324,7 @@ class Int3c2eOpt:
         if batch_size is None:
             ksh_block_partitions = [0, ksh_blocks]
         else:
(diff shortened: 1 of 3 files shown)


gpu4pyscf/df/int3c2e_bdiv.py
```diff
@@ -486,3 +570,26 @@ def estimate_shl_ovlp(mol):
     fac_norm = cs[:,None]*cs * (np.pi/aij)**1.5
     ovlp = fac_norm * cp.exp(-theta*dr**2) * fac_dri * fac_drj
     return ovlp
+
+def _parition(offsets, block_size):
```
wxj6000: _partition?

gpu4pyscf/lib/gint-rys/unrolled_int3c2e_bdiv.cu
wxj6000, line 1: Why is the header removed?

PR #423, opened 2025-06-03
Raman spectrum with PCM response off

henryw7:
image

### DIFF
--- a/gpu4pyscf/solvent/grad/pcm.py
+++ b/gpu4pyscf/solvent/grad/pcm.py
@@ -216,8 +216,8 @@ def grad_nuc(pcmobj, dm, q_sym = None):
     grid_coords  = pcmobj.surface['grid_coords'].get()
     exponents    = pcmobj.surface['charge_exp'].get()
 
-    if pcmobj.frozen_right_dm0 is not None:
-        # Note: The q_sym computed above actually use frozen_right_dm0 as input, so it's actually q_sym_right
+    if pcmobj.frozen_dm0_for_finite_difference_without_response is not None:
+        # Note: The q_sym computed above actually use frozen_dm0 as input, so it's actually q_sym_right
         q_sym_left, _ = pcmobj._get_qsym(dm, with_nuc = True)
         q_sym_left = q_sym_left.get()
         q_sym += q_sym_left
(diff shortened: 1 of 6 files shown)


sunqm:
The second equation for unperturbed density is probably incorrect. Rewriting PCM energy to
$E_{PCM} = E(D) +  E(D, q_{nuc}) +  E(q_{nuc})$, where
$E(D) = 1/2 D \cdot I(K^{-1}R)I \cdot D$ is the coupling between density matrices. $E(D,q)$ and $E(q)$ are the remaining couplings between density matrices and nulcear charges. The derivatives of PCM energy is
```math
\frac{dE(D)}{dX} = \frac{1}{2} D^0 \cdot \frac{\partial (I(KR)I)}{\partial X} \cdot D^0 + \frac{1}{2} D^0 \cdot (IKRI)^0 \cdot D^X + \frac{1}{2} D^X \cdot (IKRI)^0 \cdot D^0
```
The factor 1/2 should be there.

In a finite difference computation, E(x+dx) and E(x-dx) would need to be evaluated using this equation, where $D^0$ is frozen and it is obtained from SCF at E(x). The existing code for frozen $D^0$ only supports the first plus the third term in the above equation, which is of cause incorrect for the finite difference evaluation of the E(PCM) for a perturber. However, the second equation you listed assumes the second and the third term in the above equation is identical, which I believe is also incorrect.

gpu4pyscf/solvent/grad/pcm.py
```diff
@@ -434,6 +465,10 @@ def contract_ket(a, B, c):
 
     else:
         raise RuntimeError(f"Unknown implicit solvent model: {pcmobj.method}")
+
+    if pcmobj.frozen_dm0_for_finite_difference_without_response is not None:
+        de *= 2
```
sunqm, line 471: Add a comment why a *2 is applied here, maybe cite the discussion in this PR.

gpu4pyscf/solvent/pcm.py
```diff
@@ -425,7 +432,7 @@ def _get_qsym(self, dms, with_nuc = False):
 
         return q_sym[0], q[0]
 
-    def _get_vgrids(self, dms, with_nuc = False):
+    def _get_vgrids(self, dms, with_nuc = False, ndim = 1):
```
sunqm: The ndim=1 option is a confusing. Consider to always return 2d array. Update the tddft/pcm (or any other relevant code) to access the first vector.

gpu4pyscf/solvent/pcm.py
```diff
@@ -377,44 +379,49 @@ def _get_vind(self, dms):
...
+            # when G is the electric field, I3c, K, R doesn't depend on G, so
+            # dE^PCM/dG = dD/dG @ I3c @ (0.5*(K^-1 @ R + R^T @ (K^-1)^T)) @ I3c @ D
+            # when we are trying to take the numerical derivative of the above expression,
+            # and we want to remove the PCM response from electric field,
+            # E^PCM(dG) = D(dG) @ I3c @ (0.5*(K^-1 @ R + R^T @ (K^-1)^T)) @ I3c @ D(0)
+            # where D(dG) is the perturbed density, optimized during the perturbed SCF,
+            # we sometimes call it left dm,
+            # and D(0) is the unperturbed density, not optimized in the perturbed SCF,
+            # we call it frozen dm0 or right dm0.
+            # Compared to the original E^{PCM}, the factor of 0.5 disappears.
```
sunqm, line 413: Also cite this PR in the comment.

sunqm (approved):

PR #427, opened 2025-06-11
Optimize GPU memory usage

sunqm:
* Introduce a custom cupy memory allocator
* Optimize small-array allocations in several modules (e.g., numint) by preallocating buffers
* Cache sparse indices on the host to reduce GPU memory fragments

### DIFF
--- a/gpu4pyscf/__config__.py
+++ b/gpu4pyscf/__config__.py
@@ -38,6 +38,10 @@ if props['totalGlobalMem'] < 16 * GB:
 mem_fraction = 0.9
 cupy.get_default_memory_pool().set_limit(fraction=mem_fraction)
 
+# Threshold (in bytes) to control the custom memory allocator. The custom memory
+# allocator only cache the small-sized memory blocks in the memory pool.
+mempool_threshold = 0
+
 if props['sharedMemPerBlockOptin'] > 65536:
     shm_size = props['sharedMemPerBlockOptin']
 else:
--- a/gpu4pyscf/__config__.py
+++ b/gpu4pyscf/__config__.py
@@ -25,7 +25,7 @@ for device_id in range(num_devices):
 props = cupy.cuda.runtime.getDeviceProperties(0)
 GB = 1024*1024*1024
 min_ao_blksize = 256        # maxisum batch size of AOs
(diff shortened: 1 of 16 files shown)


gpu4pyscf/__config__.py
```diff
@@ -25,7 +25,7 @@
 props = cupy.cuda.runtime.getDeviceProperties(0)
 GB = 1024*1024*1024
 min_ao_blksize = 256        # maxisum batch size of AOs
-min_grid_blksize = 128*128  # maximum batch size of grids for DFT
+min_grid_blksize = 64*64    # maximum batch size of grids for DFT
```
wxj6000, line 28: This change most likely will slow down the numint modules, and fail the benchmark tasks.

gpu4pyscf/dft/numint.py
```diff
@@ -992,12 +1026,6 @@ def get_rho(ni, mol, dm, grids, max_memory=2000, verbose=None):
         dm = cupy.asarray(dm)
         dm = opt.sort_orbitals(dm, axis=[0,1])
 
-    mem_avail = get_avail_mem()
```
wxj6000, line 995: The dynamic block size was to balance the memory footprint and performance. 
- When the block size is small, the GPU occupation is low. Therefore the efficiency is low. Python overhead is significant.  
- On the other hand, a large block size may decrease the sparsity. If there are too many nonzero AOs, the AO block cannot fit in the memory.

gpu4pyscf/__config__.py
sunqm: A smaller batch size here to reduce memory footprint. Small batch can improve the utilization of sparsity, potentially resulting in better performance. Regarding to the performance benchmarks, minor performance fluctuations caused by adjustment like this are not worth too much attention.

gpu4pyscf/dft/numint.py
sunqm: Subtle adjustment for performance should not be considered too early. A fixed batch size is reasonable in this case.

sunqm:
I understand the concern. Minor performance changes are normal and acceptable. It's not worth to add complexity to the PR workflow

Issue #428, opened 2025-06-13
xc='HF' leads to ValueError in eval_rho(numint.py) due to unexpected ao.shape


zheliu17:
Although it might be a little bit strange to set xc='HF', GPU4PySCF fails in this case. Pure CPU version is OK.

```
import cupy
import pyscf
import gpu4pyscf

# version 1.4.1
print(gpu4pyscf.__version__)

mol = pyscf.M(
    atom="""
H   0.0 0.0 0.0
H   0.0 0.0 1.0""",
    basis="sto-3g",
)
xc = "HF"
mf = pyscf.dft.KS(mol, xc=xc).to_gpu()

mf.kernel()
```

The error message is 

```
File ****gpu4pyscf/dft/numint.py:143, in eval_rho(mol, ao, dm, non0tab, xctype, hermi, with_lapl, verbose)
    141 xctype = xctype.upper()
    142 if xctype in ('LDA', 'HF'):
--> 143     _, ngrids = ao.shape
    144 else:
    145     _, ngrids = ao[0].shape

ValueError: too many values to unpack (expected 2)
```

wxj6000:
Thank you for the feedback. The consistency should be honored. The bug has been fixed in https://github.com/pyscf/gpu4pyscf/pull/430

PR #430, opened 2025-06-14
Support xc='hf'

wxj6000:


wxj6000:
Fixing https://github.com/pyscf/gpu4pyscf/issues/428

gpu4pyscf/dft/numint.py
```diff
@@ -431,7 +431,7 @@ def _nr_rks_task(ni, mol, grids, xc_code, dm, mo_coeff, mo_occ,
         opt = ni.gdftopt
         _sorted_mol = opt._sorted_mol
         nao = _sorted_mol.nao
-        if xctype == 'LDA':
```
sunqm, line 449: Consider to return early when xctype==HF. It's unnecessary to go through the eval_rho, eval_mat for this case

gpu4pyscf/dft/numint.py
```diff
@@ -828,7 +831,7 @@ def _nr_uks_task(ni, mol, grids, xc_code, dms, mo_coeff, mo_occ,
         vmata = cupy.zeros((nset, nao, nao))
         vmatb = cupy.zeros((nset, nao, nao))
 
-        if xctype == 'LDA':
```
sunqm, line 863: return early

gpu4pyscf/dft/numint.py
```diff
@@ -1209,6 +1214,8 @@ def _nr_uks_fxc_task(ni, mol, grids, xc_code, fxc, dms, mo1, occ_coeff,
                                                   max_memory=None,
                                                   grid_range=(grid_start, grid_end)):
 
+            if xctype == 'HF':
```
sunqm, line 1248: return early

gpu4pyscf/dft/numint.py
wxj6000: I am not sure if it is proper to return early. This function also return `nelec`. It needs `eval_rho` for obtaining a meaningful value. And eval_mat is already skip in the loop.

sunqm (approved):

PR #431, opened 2025-06-17
Add gradient for tddft-ris

puzhichen:
* add TDDFT-ris gradient
* Add get_ab method for TDDFT-ris


gpu4pyscf/grad/tdrks_ris.py
```diff
@@ -0,0 +1,365 @@
...
+        if atmlst is None:
+            atmlst = self.atmlst
+        else:
+            self.atmlst = atmlst
+
+        if self.verbose >= logger.WARN:
+            self.check_sanity()
+        if self.verbose >= logger.INFO:
+            self.dump_flags()
+        theta = self.base.theta
```
sunqm, line 360: print a warning message for this experimental feature.

gpu4pyscf/df/grad/tdrhf.py
```diff
@@ -165,6 +165,120 @@ def get_jk(mf_grad, mol=None, dm0=None, hermi=0, with_j=True, with_k=True, omega
     return ej, ek, ejaux, ekaux
 
 
+def get_jk_ris(mf, mol=None, dm0=None, hermi=0, with_j=True, with_k=True, omega=None):
```
sunqm: Isn't this function identical to the get_jk function defined previously?

Does it work to introduce a separated with_df attribute for  the TD classes? However, this may require the JK derivatives to be computed separately for the ground state density matrix and the excited state density matrices separately using two with_df  instances. If this involves too many changes to the get_jk function, it can be considered in another PR.

gpu4pyscf/grad/tdrks_ris.py
```diff
@@ -0,0 +1,365 @@
...
+        else:
+            veff = 0
+        veff0mop = reduce(cp.dot, (mo_coeff.T, veff, mo_coeff))
+        wvo -= contract("ki,ai->ak", veff0mop[:nocc, :nocc], xpy) * 2
+        wvo += contract("ac,ai->ci", veff0mop[nocc:, nocc:], xpy) * 2
+        veff0mom = cp.zeros((nmo, nmo))
+
+    # set singlet=None, generate function for CPHF type response kernel
+    # TODO: LR-PCM TDDFT
+    vresp = td_grad.base._scf.gen_response(singlet=None, hermi=1)
```
sunqm, line 168: * The new API td_grad.base.gen_response is available.

* If hasattr(td_grad.base, 'with_solvent'): raise

gpu4pyscf/grad/tdrks_ris.py
```diff
@@ -0,0 +1,365 @@
...
+    return de.get()
+
+
+def get_extra_force(atom_id, envs):
+    return envs['dvhf'].aux[atom_id]
+
+def get_veff_ris(mf_J, mf_K, mol=None, dm=None, j_factor=1.0, k_factor=1.0, omega=0.0, hermi=0, verbose=None):
+    
+    if omega != 0.0:
+        vj, _, vjaux, _ = tdrhf_df.get_jk_ris(mf_J, mol, dm, omega=omega, hermi=hermi)
```
sunqm: Are the derivatives of the ground state also evaluated within the get_jk_ris using the auxiliary basis sets defined in mf_J and mf_K ? These auxbasis might not be sufficient for ground state derivatives.

puzhichen: No, get_veff_ris only calculates the contraction between ris integrals with densities from $x+y$ and $x-y$, i.e., $\langle \boldsymbol{g}^{ris(\xi)}; \boldsymbol{\Gamma}_{II} \rangle$. The contraction between ground state density matrix and normal integrals are calculated using get_veff from TDRHF.

sunqm (approved):

PR #433, opened 2025-06-17
Fix 8-fold permutation symmetry bug in j-engine

sunqm:


wxj6000:
I am still on vacation, and away from laptop. Will review the PRs from
tomorrow.

Regards,
Xiaojie

wxj6000 (approved):

PR #434, opened 2025-06-19
Introduction of _krylov_family.py for Unified Krylov Framework

John-zzh:
- This PR introduces:
 _krylov_family.py, which provides a unified Krylov-based solver framework for the following problems:

1.  Eigenvalue Problem: Solve $AX = X\Omega$, returning the eigenvalues $\Omega$  and eigenvectors $X$.
2. Linear System: Solve $AX = \text{rhs}$, returning the solution vectors $X$. rhs for right-hand-side. 
3. Shifted Linear System: Solve $AX - X\Omega = \text{rhs}$, where $\Omega$ is a diagonal matrix, returning the solution vectors $X$.

These problems share structural similarities that allow them to be addressed efficiently within a unified Krylov framework, leveraging common algorithmic components for improved performance and maintainability.

- Key Motivation: Nested Krylov Framework

The framework is designed to support a nested Krylov approach, enabling:
Non-diagonal initial guesses and non-diagonal preconditioners, solved iteratively using Krylov method.
which can accelerate Krylov convergence.

- note

1.  _krylov_family.py only depends on math_helper.py in this repo
2. The full nested Krylov functionality is not yet available in this PR, as ris.py does not currently expose the necessary matrix-vector product API. This feature will be implemented in a subsequent PR.

### DIFF
--- a/gpu4pyscf/tdscf/_krylov_family.py
+++ b/gpu4pyscf/tdscf/_krylov_family.py
@@ -1,0 +1,619 @@
+# Copyright 2021-2025 The PySCF Developers. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import cupy as cp
+import sys
+import scipy.linalg
+
+from gpu4pyscf.tdscf import math_helper
+from gpu4pyscf.lib import logger
+from functools import partial
+from pyscf.data.nist import HARTREE2EV
+
+
+def _time_add(log, t_total, t_start):
+    ''' t_total: list
+        t_start: tuple
+
+        In-place revise t_total, add the time elapsed since t_start
+    '''
+    current_t = log.timer_silent(*t_start)
+    for i, val in enumerate(current_t):
+        t_total[i] += val
+
+def _time_profiling(log, t_mvp, t_subgen, t_solve_sub, t_sub2full, t_precond, t_fill_holder, t_total):
+    '''
+    Timing breakdown:
+                            CPU(sec)  wall(sec)    GPU(ms) | Percentage 
+    mat vec product            3.61       3.67    8035.68    42.9   42.7   93.6
+    proj subspace              0.00       0.00       3.88     0.0    0.0    0.0
+    solve subspace             0.00       0.00       3.24     0.0    0.0    0.0
+    proj fullspace             0.00       0.00       1.56     0.0    0.0    0.0
+    precondition               0.37       0.43     428.91     4.4    5.0    5.0
+    fill holder                0.01       0.01       9.50     0.1    0.1    0.1
+    Sum                        4.00       4.11    8482.78    47.5   47.9   98.8
+    Total                      8.42       8.59    8587.89   100.0  100.0  100.0
+    '''
+    time_labels = ["CPU(sec)", "wall(sec)", "GPU(ms)"]
+    labels = time_labels[:len(t_total)]
+
+    log.info("Timing breakdown:")
+    header_time = " ".join(f"{label:>10}" for label in labels)
+    log.info(f"{'':<20}  {header_time} | Percentage ")
+
+    t_sum = [t_mvp[i] + t_subgen[i] + t_solve_sub[i] + t_sub2full[i] + t_precond[i] + t_fill_holder[i] for i in range(len(t_total))] 
+
+    # 计算并打印各计时器
+    timers = {
+        'mat vec product':t_mvp,
+        'proj subspace':  t_subgen,
+        'solve subspace': t_solve_sub,
+        'proj fullspace': t_sub2full,
+        'precondition':   t_precond,
+        'fill holder':    t_fill_holder,
+        'Sum':            t_sum,
+        'Total':          t_total
+    }
+    for entry, cost in timers.items():
+        time_str = " ".join(f"{x:>10.2f}" for x in cost)
+        percent_str = " ".join(f"{(x/y*100 if y != .0 else 100):>6.1f}" for x, y in zip(cost, t_total))
+        log.info(f"{entry:<20} {time_str}  {percent_str}")
+
+
+
+def eigenvalue_diagonal(**kwargs):
+    """solve
+        DX=XΩ 
+        D is diagonal matrix
+    """
+    n_states = kwargs['n_states']
+    hdiag = kwargs['hdiag']
+
+    hdiag = hdiag.reshape(-1,)
+    A_size = hdiag.shape[0]
+    Dsort = hdiag.argsort()[:n_states]
+    X = cp.zeros((n_states, A_size))
+    X[cp.arange(n_states), Dsort] = 1.0
+    return X
+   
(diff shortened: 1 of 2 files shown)


gpu4pyscf/tdscf/_krylov_family.py
```diff
@@ -0,0 +1,619 @@
...
+        rhs=rhs, omega_shift=omega_shift, extra_init=extra_init,
+        initguess_fn=initguess_fn, precond_fn=precond_fn,
+        conv_tol=conv_tol, max_iter=max_iter, 
+        gram_schmidt=gram_schmidt, single=single, verbose=verbose
+    )
+
+    return output
+
+
+def test_krylov_solver():
```
sunqm: move this test to a separated test file.

gpu4pyscf/tdscf/_krylov_family.py
```diff
@@ -0,0 +1,619 @@
...
+        initguess_fn = globals()[f'_{problem_type}_diagonal_initguess']
+
+
+    ''' Generate initial guess '''
+    log.info('generating initial guess')
+    if problem_type == 'eigenvalue':
+        init_guess = initguess_fn(n_states=size_new, hdiag=hdiag)
+        if isinstance(init_guess, tuple):
+            energies, init_guess_vec = init_guess
+            energies = energies*HARTREE2EV
```
sunqm: Using AU internally

gpu4pyscf/tdscf/_krylov_family.py
```diff
@@ -0,0 +1,619 @@
...
+        '''
+        log.info(nks_citation)
+        fill_holder = math_helper.nKs_fill_holder
+        s_holder = cp.empty_like(sub_A_holder)
+
+    if initguess_fn and callable(initguess_fn):
+        log.info(' use user-specified function to generate initial guess.')
+    else:
+        log.info(' use hdiag to generate initial guess.')
+        initguess_fn = globals()[f'_{problem_type}_diagonal_initguess']
```
sunqm: Using `if...elif...` or lookup table to explicitly assign the initguess_fn

gpu4pyscf/tdscf/_krylov_family.py
```diff
@@ -0,0 +1,619 @@
...
+    
+    V_holder, size_new = fill_holder(V_holder, size_old, init_guess)
+    log.info('initial guess done')
+    
+
+    if precond_fn and callable(precond_fn):
+        log.info(' use user-specified function for preconditioning.')
+    else:
+        log.info(' use hdiag for preconditioning.')   
+        precond_fn = globals()[f'_{problem_type}_diagonal_precond']
```
sunqm: Explicit assignment for precond_fn

gpu4pyscf/tdscf/_krylov_family.py
```diff
@@ -0,0 +1,594 @@
...
+            overlap_s = s_holder[:size_new, :size_new]
+
+        if problem_type == 'eigenvalue':
+            if gram_schmidt:
+                ''' solve ax=xΩ '''
+                omega, x = cp.linalg.eigh(sub_A)
+            else:
+                ''' solve ax=sxΩ 
+                TODO need precondition step: s/d first'''
+                omega, x = scipy.linalg.eigh(sub_A.get(), overlap_s.get())
```
wxj6000: Why do you need to call eigen solver on CPU? We have the generalized eigen solver in cupy_helpy.

gpu4pyscf/tdscf/tests/test_krylov.py
```diff
@@ -0,0 +1,132 @@
...
+            omega_shift=self.omega_shift,
+            conv_tol=1e-12,
+            max_iter=35,
+            gram_schmidt=True,
+            verbose=4,
+            single=False
+        )
+
+        # Compare solutions
+        self.assertAlmostEqual(
```
wxj6000, line 196: Add test for single precision

gpu4pyscf/tdscf/_krylov_family.py
```diff
@@ -0,0 +1,594 @@
...
+    time_labels = ["CPU(sec)", "wall(sec)", "GPU(ms)"]
+    labels = time_labels[:len(t_total)]
+
+    log.info("Timing breakdown:")
+    header_time = " ".join(f"{label:>10}" for label in labels)
+    log.info(f"{'':<20}  {header_time} | Percentage ")
+
+    t_sum = [t_mvp[i] + t_subgen[i] + t_solve_sub[i] + t_sub2full[i] + t_precond[i] + t_fill_holder[i] for i in range(len(t_total))] 
+
+    # 计算并打印各计时器
```
wxj6000: Translate it in English for readability.

John-zzh:
> The Krylov subspace family also includes CG, GMRES, and so on. The unified framework is a bit too general. And I suggest
> 
> * Split solving the eigenvalue problem and solving linear system into two functions. Linear system and shifted linear system probably can be merged. aka, agreed with Qiming's comment.
> * In krylov_solver, simply take initial guess (value), preconditioner (value), dot_product (function), Gram-Schmidt orthogonalization or not (bool), and other controls as the input. Not generated functions. This function probably can be used in other routines.
> * In nested_krylov_solver, generate initial guess, preconditioner, and dot_product explicitly.

Thank you for pointing out the broad scope of the Krylov subspace family. You're right that the term "family" might be too general. A more specific name like "Krylov subspace tools" could better reflect the scope of the current implementation, which is not intended to encompass all methods like CG or GMRES.

The all-in-one design is driven by the need to support iterative preconditioning (nested Krylov) while maximizing code reuse, particularly for linear response problems. Splitting the eigenvalue and linear system solvers into separate functions would necessitate splitting nested_krylov_solver into nested_krylov_eigensolver and nested_krylov_linear_solver, leading to significant code duplication. Furthermore, restricting the preconditioner in krylov_solver to a static value would eliminate the ability to perform iterative preconditioning, such as ris-preconditioning, which is the core feature of the current design.

I understand that the design may seem unconventional at first glance (and perhaps even more complex upon deeper inspection :-) ). However, it remains manageable and is not more complicated than recursive ERI computations.

gpu4pyscf/tdscf/_krylov_family.py
John-zzh: ahahaha, because I didn't dig around the valuable infrastructure. 
Do you mean  this?https://github.com/pyscf/gpu4pyscf/blob/a00e0bab734decda6e39dcf1726b094642eeb4a7/gpu4pyscf/lib/cusolver.py#L101

wxj6000: Yes

John-zzh: it only supports np.float64, np.complex128 though. I also need float32

John-zzh:
I tried several wrappers that explicitly split the eigenvalue, linear and shifted_linear problems. 
for example, the wrapper for eigenvalue is like
```
def krylov_eigenvalue(matrix_vector_product, hdiag, 
                  initguess_fn=None, precond_fn=None, 
                  n_states=20,conv_tol=1e-5, 
                  max_iter=35, extra_init=8, gram_schmidt=True, 
                  single=False, verbose=logger.NOTE):
    
    eigenvalue, eigenvector = krylov_solver(matrix_vector_product=matrix_vector_product, 
                   hdiag=hdiag, problem_type='eigenvalue', 
                  initguess_fn=initguess_fn, precond_fn=precond_fn, n_states=n_states,conv_tol=conv_tol, 
                  max_iter=max_iter, extra_init=extra_init, gram_schmidt=gram_schmidt, 
                  single=single, verbose=verbose)
    return eigenvalue, eigenvector 
```
looks pretty tedious though. Only the "problem_type" need to be explicitly assigned here, other args are simply parsing to krylov_solver.  Even using **kwargs, we still need triple  amount duplicated doc string for each arg.

 I think for now the current implementation is fine, users just need to explicitly specify the 'problem_type'

PR #443, opened 2025-07-01
Add geometry optimization for excited states using TDDFT-ris

puzhichen:


gpu4pyscf/grad/tdrks_ris.py
```diff
@@ -319,6 +317,67 @@ def get_veff_ris(mf_J, mf_K, mol=None, dm=None, j_factor=1.0, k_factor=1.0, omeg
     return vhf
 
 
+def as_scanner(td_grad, state=1):
```
sunqm: This function is likely identical to the one defined in the tdrks.

gpu4pyscf/tdscf/ris.py
```diff
@@ -120,13 +120,8 @@ def get_auxmol(mol, theta=0.2, fitting_basis='s'):
...
     '''
-    auxmol = gto.M(atom=mol.atom,
-                    basis=mol.basis,
-                    parse_arg=False,
-                    spin=mol.spin,
-                    charge=mol.charge,
-                    cart=mol.cart)
-
+    auxmol = mol.copy()
+    auxmol.parse_arg = False
```
sunqm: parse_arg should be passed through the build() function. Actually, this argument is not necessary as the default setting has been changed to `False` since pyscf-2.7

gpu4pyscf/tdscf/ris.py
```diff
@@ -700,6 +695,36 @@ def add_hf_(a, b, hyb=1):
 
     return a.get(), b.get()
 
+
+def as_scanner(td):
```
sunqm, line 697: it's identical to the one defined in tdrks.

sunqm (approved):

PR #460, opened 2025-07-17
In EDA, give an error if the output will likely be lost when redirected to one file

henryw7:


gpu4pyscf/properties/eda.py
```diff
@@ -1092,9 +1094,23 @@ def _get_gradient(mf):
...
         return grad_obj.kernel()
 
     n_frag = len(mol_list)
+    for i_frag in range(n_frag):
+        for j_frag in range(i_frag + 1, n_frag):
+            if mol_list[i_frag].stdout != mol_list[j_frag].stdout:
+                warnings.warn("The stdout of each mol in mol_list is not consistent. We do not guarantee which stdout to write.")
+                if mol_list[i_frag].stdout.name == mol_list[j_frag].stdout.name:
+                    import os
+                    if mol_list[i_frag].stdout.name != os.devnull:
```
sunqm: `.name` is unsafe. `stdout` can be a string buffer object (IOString) . It does not have the .name attribute . I think this error check is unnecessary

sunqm (approved):

sunqm (approved):

PR #470, opened 2025-08-06
Add minimum energy crossing point search function

puzhichen:
* Add the function to get the minimum energy crossing point.
* Fix a bug in finite difference of NACV.


gpu4pyscf/nac/mecp.py
```diff
@@ -0,0 +1,243 @@
...
+        self.verbose = self.td.verbose
+        self.crossing_type = crossing_type
+        self.stdout = self.td.stdout
+        self.log = logger.new_logger(self, self.verbose)
+        
+        self._last_geom = None
+        self._last_energy = None
+        self._last_grad = None
+
+    class _GradScanner:
```
sunqm: Move this class definition to module level

gpu4pyscf/nac/mecp.py
```diff
@@ -0,0 +1,243 @@
...
+            """Returns the scanner object."""
+            return self
+
+        def converged(self):
+            td_scanner = self.base
+            return all((td_scanner._scf.converged,
+                        td_scanner.converged[self.states[0]-1],
+                        td_scanner.converged[self.states[1]-1]))
+
+    class _SolverWrapper:
```
sunqm: To module level

gpu4pyscf/nac/mecp.py
```diff
@@ -0,0 +1,243 @@
...
+        def nuc_grad_method(self):
+            return self.ci_optimizer._GradScanner(self)
+
+
+    def get_eff_energy_and_gradient(self):
+        current_geom = self.mol.atom_coords(unit='bohr')
+
+        self.log.info("\n--- CI Optimizer Step (using pyscf.geomopt.geometric_solver) ---")
+
+        self.mf.mol = self.mol
```
sunqm: These inplace updates are unsafe:
* It implicitly changes the `mol` attributes of the underlying mf and td objects.  However, certain attributes (such as .with_df, ._eri, ) are not updated consistently
* The mo_coeff for mf and xy for td methods from previous steps may or may not be used as the initial guess for the current step, this depend on the individual method implementation. However, in the geometry optimization we always want to utilize the initial guess from previous calculations.

It's recommended to call the scanner during the optimization. It can be a `td_scanner = td.as_scanner()` for this optimization solver. Calling the `td_scanner(new_geometry)` will clean up the relevant intermediates. In the output, the xy amplitudes of td_scanner and the mo_coeff, mo_energy of `td_scanner._scf` are updated. They can be used to compute the NACV or gradients.

gpu4pyscf/nac/mecp.py
```diff
@@ -0,0 +1,243 @@
...
+        self.td.mol = self.mol
+
+        self.mf.kernel()
```
sunqm: These kernel may not be able to correctly apply initial guess from previous calculations

gpu4pyscf/nac/mecp.py
```diff
@@ -0,0 +1,243 @@
...
+                                      from the molecule in the MCSCF object.
+            **kwargs: Additional keyword arguments to pass to the optimizer.
+                      e.g., max_cycle=100, prefix='traj.xyz'
+            TODO: kwargs seem not work!
+        """
+        if geom is not None:
+            self.mol.atom = geom
+
+        # Create the wrapper object that geometric_solver will use
+        solver_wrapper = self._SolverWrapper(self)
```
sunqm: the solver_wrapper can be created as a pyscf.lib.GradScanner instance. The geometric_solver function will execute this as `solver_wrapper(new_goemetry)` . The `td_scanner` mentioned in the `get_eff_energy_and_gradient` function can be called in the scanner wrapper.

Adaptation to ris can be executed in `solver_wrapper()`. The mean-field object in ris can be created as `mf.as_scanner()`. and mf calculation should be executed through the mf_scanner. A `reset` method for ris should be added, to cleanup intermediates for the old geometry. and the ris' reset() should also be called within the `solver_wrapper()`

sunqm (approved):

PR #476, opened 2025-08-08
Multigrid v2 with MGGA

henryw7:


gpu4pyscf/pbc/dft/multigrid_v2.py
```diff
@@ -459,8 +459,11 @@ def sort_gaussian_pairs(mydf, xc_type="LDA"):
     return mydf
 
 
-def evaluate_density_wrapper(pairs_info, dm_slice, img_phase, ignore_imag=True):
-    c_driver = libgpbc.evaluate_density_driver
+def evaluate_density_wrapper(pairs_info, dm_slice, img_phase, ignore_imag=True, compute_tau=False):
+    if compute_tau:
```
Walter-Feng, line 463: Just had a glance, looks like this flag would be slightly confusing. I would imagine "if we compute tau, we use evaluate_density_tau_driver".

henryw7: Fixed, nice catch

gpu4pyscf/pbc/dft/multigrid_v2.py
```diff
@@ -612,7 +637,13 @@ def evaluate_density_on_g_mesh(mydf, dm_kpts, kpts=None, xc_type='LDA'):
 
 
 def evaluate_xc_wrapper(pairs_info, xc_weights, img_phase):
```
sunqm, line 639: pass flag like `with_tau` or `xctype` to this function. and check the dimension of xc_weights accordingly.

gpu4pyscf/pbc/dft/multigrid_v2.py
```diff
@@ -686,8 +717,16 @@ def convert_xc_on_g_mesh_to_fock(
     cell = mydf.cell
     nao = cell.nao_nr()
 
-    xc_on_g_mesh = xc_on_g_mesh.reshape(-1, *mydf.mesh)
+    # TODO: This logic will cause bugs with kpts
+    if xc_on_g_mesh.ndim < 3: # 1 for n_channels, 1 for rho/tau, 1 for ngrids
```
sunqm, line 721: A little bit misleading and error-prone. `density_slices` can be determined based on the `with_tau/xctype` flag. then reshape can be performed as `xc_on_g_mesh.reshape(-1, density_slices, *mesh)`

sunqm (approved):

PR #489, opened 2025-08-23
Improve numerical stability for multigrid v2

Walter-Feng:
This PR aims to (slightly) improve numerical stability for the recursion relation of gaussian pairs in multigrid v2 by the following two changes:

1. screen out the pairs at blocks where the recursion factors go too large
2. distribute the pair_prefactor into the components of gaussian

The screened subblock of pairs should be re-calculated via explicit gaussian calculation instead of RR, which should be introduced in later PRs.

Note that a lot of actual changes introduced in this PR is mainly due to the format changes - it would be helpful to have a `.clang-format` file fixed for all contributors.

### DIFF
--- a/gpu4pyscf/lib/multigrid/multigrid_v2/gradient.cuh
+++ b/gpu4pyscf/lib/multigrid/multigrid_v2/gradient.cuh
@@ -78,15 +78,15 @@ __global__ void evaluate_xc_kernel(
   const KernelType start_position_z =
       dxyz_dabc[2] * a_start + dxyz_dabc[5] * b_start + dxyz_dabc[8] * c_start;
 
-  const KernelType a_dot_b = dxyz_dabc[0] * dxyz_dabc[3]
-                           + dxyz_dabc[1] * dxyz_dabc[4]
-                           + dxyz_dabc[2] * dxyz_dabc[5];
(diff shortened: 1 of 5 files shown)


gpu4pyscf/lib/multigrid/multigrid_v2/gradient.cuh
```diff
@@ -268,41 +271,42 @@ __global__ void evaluate_xc_kernel(
     } else {
       x = start_position_x;
     }
-    for (a_index = 0, gaussian_x = 1,
-         recursion_factor_a = recursion_factor_a_start;
-         a_index < a_upper;
-         a_index++, gaussian_x *= recursion_factor_a,
-         recursion_factor_a *= exp_da_squared) {
+    for (a_index = 0, gaussian_x = gaussian_starting_point,
+        recursion_factor_a = recursion_factor_a_start;
```
sunqm, line 275: Is it possible that `recursion_factor_a_start` is inf?

Walter-Feng: Yes, generally that is the source of `inf`.  My way to address that is removing such pairs/regions in the prescreening step, so that in actual calculations `recursion_factor_a_start` remains within double precision range (< 10^173).

PR #496, opened 2025-08-29
Multigrid v2 with k point support

henryw7:


gpu4pyscf/pbc/grad/tests/test_pbc_grad_krks.py
```diff
@@ -14,10 +14,11 @@
 # limitations under the License.
 
 import unittest
-from pyscf import lib
+import numpy as np
 from pyscf.pbc import gto
+from gpu4pyscf.pbc.dft import multigrid_v2
 
-disp = 1e-3
+disp = 1e-5
```
sunqm, line 21: This is too small. The integral evaluation cannot achieve very high accuracy. The energy computation may lead to more uncertainty.

sunqm (approved):

gpu4pyscf/pbc/grad/tests/test_pbc_grad_krks.py
henryw7: For a small and simple system like the diamond primitive cell with a less-than-minimal basis set, dx=1e-5 is achievable, and actually matches analytical gradient very well.

PR #522, opened 2025-09-24
Fermi smearing for molecular case

Walter-Feng:
This PR aims to add basic support for Fermi smearing in molecular systems, which addresses an urgent request. 

Numerically it keeps exactly the same with PySCF routines. The only changes involved are data conversion among scipy, numpy and cupy. Nevertheless, to speed up implementation, I skipped a lot of relativistic stuffs and spins that may involve some `intor`s.

I only tested with `00-h20.py` in the examples, which shows it is compatible with the basic routines such as energy, gradient and hessian. I am also open to implementing any numerical unit tests.

### DIFF
--- a/gpu4pyscf/scf/addons.py
+++ b/gpu4pyscf/scf/addons.py
@@ -16,13 +16,33 @@ from functools import reduce
 
 import cupy
 import scipy
-from pyscf import __config__, lib
-from pyscf.pbc.tools import print_mo_energy_occ
 
 from gpu4pyscf import scf
 from gpu4pyscf.lib import logger
+from pyscf import __config__, lib
 
 SMEARING_METHOD = getattr(__config__, "pbc_scf_addons_smearing_method", "fermi")
+
+
+def print_mo_energy_occ(mf, mo_energy, mo_occ, is_uhf):
+
+    if is_uhf:
+        nocc = len(mo_energy[0])
+        cupy.set_printoptions(precision=6, threshold=nocc, suppress=True)
+        logger.debug(mf, "  alpha mo_energy/mo_occ")
+        logger.debug(mf, "  %s", mo_energy[0])
+        logger.debug(mf, "  %s", mo_occ[0])
+        logger.debug(mf, "  beta  mo_energy/mo_occ")
+        logger.debug(mf, "  %s", mo_energy[1])
+        logger.debug(mf, "  %s", mo_occ[1])
+        cupy.set_printoptions()
+    else:
+        nocc = len(mo_energy)
+        cupy.set_printoptions(precision=6, threshold=nocc, suppress=True)
+        logger.debug(mf, "  mo_energy/mo_occ")
+        logger.debug(mf, "  %s", mo_energy)
+        logger.debug(mf, "  %s", mo_occ)
+        cupy.set_printoptions()
 
 
 def _fermi_smearing_occ(mu, mo_energy, sigma):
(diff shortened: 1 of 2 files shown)


sunqm:
Please merge the master branch

gpu4pyscf/scf/addons.py
```diff
@@ -0,0 +1,277 @@
+from functools import reduce
```
sunqm, line 15: Add apache license header

gpu4pyscf/scf/addons.py
```diff
@@ -0,0 +1,277 @@
...
+
+
+def _smearing_optimize(f_occ, mo_es, nocc, sigma):
+    def nelec_cost_fn(m):
+        m = cupy.asarray(m)
+        mo_occ = f_occ(m, mo_es, sigma)
+        result = (mo_occ.sum() - nocc) ** 2
+        return float(result)
+
+    fermi = _get_fermi(mo_es, nocc).get()
```
sunqm, line 78: Can the PySCF CPU code be directly called here? I think that would just require converting `f_occ` and `mo_es` into np array, and cp.asarray() for mu and mo_occs in the return

sunqm:
Please add a small test

gpu4pyscf/scf/addons.py
Walter-Feng: I have copied this function to this file so it won't be calling the pyscf routines. Nevertheless, I do see some functions yet that have not been cleaned up (e.g. `print_mo_energy_occ`), so will be giving a thorough check with a test as required.

gpu4pyscf/scf/addons.py
Walter-Feng: Added

gpu4pyscf/scf/addons.py
```diff
@@ -1,9 +1,25 @@
+# Copyright 2021-2024 The PySCF Developers. All Rights Reserved.
```
wxj6000: It is already 2025

gpu4pyscf/scf/tests/test_fermi_smearing.py
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env python
```
wxj6000: Follow the unit test of this repo.

gpu4pyscf/scf/addons.py
```diff
@@ -1,9 +1,25 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 from functools import reduce
 
 import cupy
 import scipy
 from pyscf import __config__, lib
+from pyscf.pbc.tools import print_mo_energy_occ
```
wxj6000: Why does SCF module need the dependency of pbc module?

gpu4pyscf/scf/addons.py
Walter-Feng: I simply copied from the neighboring file - will modify, just a note that maybe the other file will also need to be updated.

gpu4pyscf/scf/addons.py
Walter-Feng: PySCF originally does this - I would imagine printing eigenvalues can differ between k-points and gamma-point, so they are differentiated in pbc. Maybe a historical reason, but it is only defined in pbc.

wxj6000: I feel it is better to move it out, or create a new function.

Walter-Feng: Removed dependency

gpu4pyscf/scf/tests/test_fermi_smearing.py
Walter-Feng: cleaned

gpu4pyscf/scf/addons.py
Walter-Feng: updated

PR #528, opened 2025-09-27
Support overlap matrix with large condition number

henryw7:
Not turned on by default, set `gpu4pyscf.__config__.scf_hf_remove_overlap_zero_eigenvalue = True` if needed.

### DIFF
--- a/gpu4pyscf/lib/cupy_helper.py
+++ b/gpu4pyscf/lib/cupy_helper.py
@@ -1149,3 +1149,15 @@ def set_conditional_mempool_malloc(n_bytes_threshold=100000000):
             return cuda_malloc(size)
         return default_mempool_malloc(size)
     cupy.cuda.set_allocator(malloc)
+
+# def safe_eigh(h, s, lindep=1e-6):
+#     e, v = cupy.linalg.eigh(s)
+#     x = v[:,e>lindep] / cupy.sqrt(e[e>lindep])
+#     xhx = x.T @ h @ x
+#     e, c = cupy.linalg.eigh(xhx)
+#     c = cupy.dot(x, c)
+
+#     nao, nmo = x.shape
+#     c = cupy.concatenate([c, cupy.zeros([nao, nao-nmo])], axis = 1)
+#     e = cupy.concatenate([e, cupy.zeros(nao-nmo) + 1000000])
+#     return e, c
--- a/gpu4pyscf/scf/diis.py
+++ b/gpu4pyscf/scf/diis.py
@@ -47,16 +47,20 @@ class CDIIS(lib.diis.DIIS):
         self.Corth = None
         self.space = 8
 
-    def update(self, s, d, f, *args, **kwargs):
-        errvec = self._sdf_err_vec(s, d, f)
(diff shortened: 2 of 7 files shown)


gpu4pyscf/scf/tests/test_diffuse_orbital.py
```diff
@@ -0,0 +1,222 @@
...
+            X    P
+                0.010000 1.0
+            X    P
+                0.010001 1.0
+        """,
+        verbose = 5,
+        output = '/dev/null',
+    )
+
+    # ref_mol = pyscf.M(
```
sunqm, line 46: This unused code can be removed

gpu4pyscf/lib/cupy_helper.py
```diff
@@ -1149,3 +1149,15 @@ def malloc(size):
             return cuda_malloc(size)
         return default_mempool_malloc(size)
     cupy.cuda.set_allocator(malloc)
+
```
sunqm: This unused code can be removed.

gpu4pyscf/scf/diis.py
```diff
@@ -47,16 +47,20 @@ def __init__(self, mf=None, filename=None):
         self.Corth = None
         self.space = 8
 
-    def update(self, s, d, f, *args, **kwargs):
-        errvec = self._sdf_err_vec(s, d, f)
+    def update(self, s, d, f, *args, x=None, **kwargs):
```
sunqm: `x` is too ambiguous. There is no need to pass `x` to these functions.  In the case of singular overlap matrix, Corth can be initialized in the SCF driver.

gpu4pyscf/scf/diis.py
```diff
@@ -69,13 +73,17 @@ def get_num_vec(self):
...
         '''error vector = SDF - FDS'''
         if f.ndim == s.ndim+1: # UHF
             assert len(f) == 2
             if s.ndim == 2: # molecular SCF or single k-point
                 if self.Corth is None:
-                    self.Corth = eigh(f[0], s)[1]
+                    if x is None:
+                        self.Corth = eigh(f[0], s)[1]
+                    else:
+                        assert x.ndim == 2 and x.shape[0] == s.shape[0]
```
sunqm: When `overlap_canonical_decomposed_x` is available, it can be directly assigned to the Corth when initializing the diis instance in the SCF driver, here
https://github.com/pyscf/gpu4pyscf/blob/c863f8f1e81b347f2bbaf6e5ae67a75a084de697/gpu4pyscf/scf/hf.py#L227
 and https://github.com/pyscf/gpu4pyscf/blob/c863f8f1e81b347f2bbaf6e5ae67a75a084de697/gpu4pyscf/scf/hf_lowmem.py#L94

gpu4pyscf/scf/tests/test_diffuse_orbital.py
henryw7: This ref_mol object shows how the reference value of each test is obtained, I believe it's useful for future reference.

PR #531, opened 2025-09-30
dimension mismatch in scale_ao

sunqm:


gpu4pyscf/dft/numint.py
```diff
@@ -2247,22 +2247,20 @@ def _tau_dot_sparse(bra, ket, wv, nbins, screen_index, ao_loc,
 
 def _scale_ao(ao, wv, out=None):
     if wv.ndim == 1:
-        if ao.flags.f_contiguous or ao.dtype != np.float64:
-            return cupy.multiply(ao, wv, out=out)
         nvar = 1
         nao, ngrids = ao.shape
         assert wv.size == ngrids
+        out = ndarray((nao, ngrids), dtype=ao.dtype, buffer=out)
+        if ao.flags.f_contiguous or ao.dtype != np.float64:
```
henryw7: It makes more sense to check "if not ao[0].flags.c_contiguous", because we can encounter an array both c and f contiguous (1d array)

PR #532, opened 2025-10-01
More numerically stable Hessian for orbitals with large exponent

henryw7:
- [x] RHF (Hcore second derivative)
- [ ] RKS (xc second derivative)
- [x] density fitting (auxbasis_response == 2 must be true, no problem in integral derivatives)


henryw7:
Yes it is there at https://github.com/deepprinciple/pysisyphus/blob/master/pysisyphus%2Fcalculators%2FPySCF.py#L280

henryw7:
Nice catch, probably an incorrect indentation. I'll fix it later together with other fixes.

sunqm (approved):

PR #594, opened 2025-12-11
PCM lowmem mode, without ngrids^2 storage

henryw7:
This PR is unrelated to domain decomposition PCM available in pyscf. It is just reconstructing every S, D (thus R, K^-1) matrices on the fly, similar to the logic of direct SCF. K^-1 is obtained via GMRES in cupy.

- [x] CPCM energy
- [x] IEFPCM energy
- [x] CPCM gradient
- [x] IEFPCM gradient

### DIFF
--- a/gpu4pyscf/solvent/tests/test_pcm_lowmem_grad.py
+++ b/gpu4pyscf/solvent/tests/test_pcm_lowmem_grad.py
@@ -49,6 +49,7 @@ def _gradient_with_solvent(mf, method):
     cm.verbose = 0
     cm.lebedev_order = lebedev_order
     cm.method = method
+    cm.lowmem_intermediate_storage = True
     mf = mf.PCM(cm)
     mf.conv_tol = 1e-10
     mf.kernel()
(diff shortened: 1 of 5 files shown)


gpu4pyscf/solvent/tests/test_pcm_lowmem_grad.py
```diff
@@ -73,12 +75,12 @@ def test_lowmem_gradient_RHF_CPCM(self):
         assert diff < g_tolerance
```
sunqm, line 75: You can relax the tolerance to 1e-6 or larger, than modifying the test results. Gradients errors < 1e-7 are easily broken by numerical noises and parameter adjustments.

sunqm (approved):

gpu4pyscf/solvent/tests/test_pcm_lowmem_grad.py
henryw7: For this simple H2O2 system with small 6-31g basis and small lebedev_order=11 PCM grid, 1e-7 should be achievable, and it seems stable enough among different PCM algorithms and types of GPUs.

PR #609, opened 2025-12-22
Accelerate DFT gradient grid response

henryw7:


gpu4pyscf/grad/rks.py
```diff
@@ -32,6 +32,7 @@
 from gpu4pyscf.lib import logger
 from gpu4pyscf.__config__ import num_devices
 from gpu4pyscf.dft.numint import NLC_REMOVE_ZERO_RHO_GRID_THRESHOLD
+from gpu4pyscf.hessian.rks import get_dweight_dA
```
sunqm: This import may cause circular dependency between modules. Move this import into `grids_response_cc`

sunqm (approved):

PR #615, opened 2025-12-28
TDA-ris capable of 30000 nbf (3000 atoms) 

John-zzh:
major updates compared to 1.4.1

1. 3c2e engine updated with 3c2e_bdiv
2. on the fly Coulomb iajb instead of storing tensor
3. store Davidson trial vectors in CPU mem
4. many other optimizations

As of date, some version of MKL-numpy is not compatible with cupy 13.4.1 and cutensor  2.2.0.0, contract engine will complaine about STATUS,
while openblas-numpy-scipy has segmentation fault when solving large FC=SCE.  

To use TDA-ris on large system, use
```
mamba install -c conda-forge \
  cupy=13.4.1 \
  cutensor=2.2.0 \
  numpy scipy \
  blas=*=*mkl
```

```
(cupy_mkl_py311) root@volcengine:~/jojo/software/gpu4pyscf_ris/gpu4pyscf/tdscf# ls /root/jojo/miniconda3/envs/cupy_mkl_py311/lib | grep blas
libblas.so
libblas.so.3
libcblas.so
libcblas.so.3
libcublasLt.so.12
libcublasLt.so.12.9.1.4
libcublas.so.12
libcublas.so.12.9.1.4
libnvblas.so.12
libnvblas.so.12.9.1.4
```

```
 python -c "import numpy as np; np.show_config()"
... 
  "Build Dependencies": {
    "blas": {
      "name": "blas",
      "found": true,
      "version": "3.9.0",
      "detection method": "pkgconfig",
      "include directory": "/root/jojo/miniconda3/envs/cupy_mkl_py311/include",
      "lib directory": "/root/jojo/miniconda3/envs/cupy_mkl_py311/lib",
      "openblas configuration": "unknown",
      "pc file directory": "/root/jojo/miniconda3/envs/cupy_mkl_py311/lib/pkgconfig"
    },
    "lapack": {
      "name": "lapack",
      "found": true,
      "version": "3.9.0",
      "detection method": "pkgconfig",
      "include directory": "/root/jojo/miniconda3/envs/cupy_mkl_py311/include",
      "lib directory": "/root/jojo/miniconda3/envs/cupy_mkl_py311/lib",
      "openblas configuration": "unknown",
      "pc file directory": "/root/jojo/miniconda3/envs/cupy_mkl_py311/lib/pkgconfig"
    }
...
```

### DIFF
--- a/gpu4pyscf/tdscf/ris.py
+++ b/gpu4pyscf/tdscf/ris.py
@@ -507,6 +507,7 @@ class RisBase(lib.StreamObject):
 
         if self.RKS:
             mdpol = get_inter_contract_C(int_tensor=int_rxp, C_occ=self.C_occ_notrunc, C_vir=self.C_vir_notrunc)
+
         else:
             ''' TODO '''
             mdpol_alpha = get_inter_contract_C(int_tensor=int_rxp, C_occ=self.C_occ[0], C_vir=self.C_vir[0])
@@ -920,7 +921,7 @@ def get_Tpq(mol, auxmol, lower_inv_eri2c, C_p, C_q,
 
 
     bytes_per_aux = ( n_eri3c_per_aux + n_eri3c_unzip_per_aux + n_Ppq_per_aux) * cp_int3c_dtype.itemsize
-    batch_size = min(naux, max(1, int(available_gpu_memory * 0.5 // bytes_per_aux)) )
+    batch_size = min(naux, max(1, int(available_gpu_memory * 0.4 // bytes_per_aux)) )
 
     DEBUG = False
     if DEBUG:
--- a/gpu4pyscf/tdscf/_krylov_tools_casida.py
+++ b/gpu4pyscf/tdscf/_krylov_tools_casida.py
@@ -367,6 +367,7 @@ def ABBA_krylov_solver(matrix_vector_product, hdiag, problem_type='eigenvalue',
     xp = np if in_ram else cp
     log.info(f'xp {xp}')
 
+    log.info(gpu_mem_info('before build holders'))
 
     V_p_W_holder = xp.empty((max_N_mv, A_size), dtype=hdiag.dtype)
     V_m_W_holder = xp.empty_like(V_p_W_holder)
@@ -377,6 +378,9 @@ def ABBA_krylov_solver(matrix_vector_product, hdiag, problem_type='eigenvalue',
     sub_a_p_b_holder = cp.empty((max_N_mv,max_N_mv), dtype=hdiag.dtype)
     sub_a_m_b_holder = cp.empty_like(sub_a_p_b_holder)
     sub_sigma_p_pi_holder = cp.empty_like(sub_a_p_b_holder)
+
+    log.info(gpu_mem_info('after build holders'))
+
 
     '''
     set up initial guess, V= TDA initial guess, W=0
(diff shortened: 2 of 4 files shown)


gpu4pyscf/tdscf/ris.py
```diff
@@ -803,10 +1892,12 @@ def __init__(self, mf,
...
             gram_schmidt (bool, optional): Whether to calculate the ground state. Defaults to False.
             single (bool, optional): Whether to use single precision. Defaults to True.
-            group_size (int, optional): Group size for the integral calculation. Defaults to 256.
-            group_size_aux (int, optional): Group size for the auxiliary integral calculation. Defaults to 256.
-            in_ram (bool, optional): Whether to perform calculations in RAM. Defaults to True.
+            tensor_in_ram (bool, optional): Whether to store Tpq tensors in RAM. Defaults to False.
+            krylov_in_ram (bool, optional): Whether to store Krylov vectors in RAM. Defaults to False.
             verbose (optional): Verbosity level of the logger. If None, it will use the verbosity of `mf`.
+            nto_state (None or int, optional): Which state to calculate natural transition orbitals, 
+                                        require install MOKIT  https://jeanwsr.gitlab.io/mokit-doc-mdbook/
```
jeanwsr: please use the gitlab url when suggesting installing MOKIT.

gpu4pyscf/tdscf/ris.py
```diff
@@ -745,28 +1722,140 @@ def __call__(self, mol_or_geom, **kwargs):
...
-        self.mo_coeff = cp.asarray(self._scf.mo_coeff, dtype=self.dtype)
+        self.mo_coeff = cuasarray(self._scf.mo_coeff, dtype=self.dtype)
         self.build()
         self.kernel()
         return mf_e + self.energies/HARTREE2EV
 
+def get_nto(self,state_id):
+    
+    ''' need to install MOKIT to dump .fch format orbital file
+    https://jeanwsr.gitlab.io/mokit-doc-mdbook/
```
jeanwsr: and here.

gpu4pyscf/tdscf/ris.py
```diff
@@ -745,28 +1722,140 @@ def __call__(self, mol_or_geom, **kwargs):
...
+    nto_mf.mo_coeff = nto_coeff.get()
+    nto_mf.mo_energy = cp.asarray([dominant_weight, dominant_weight]).get()
+
+    fchfilename = f'ntopair_{state_id}.fch'
+    
+    from mokit.lib.py2fch_direct import fchk
+    fchk(nto_mf, fchfilename)
+    from mokit.lib.rwwfn import del_dm_in_fch
+    del_dm_in_fch(fchname=fchfilename,itype=1)
+    return nto_mf
```
jeanwsr: I think the get_nto function should only return NTO coeffs, etc. and dump fchk can be done in other functions, to keep the code more organized. Also, in this way it still work when mokit is missing.

John-zzh: meaning dump fchk in a independent py file, not in GPU4pyscf repo?

jeanwsr: I mean, one function `get_nto` for calculating and returning NTOs only, like the api provided by [tddft](https://github.com/pyscf/pyscf/blob/master/examples/tddft/01-nto_analysis.py), and another `save_nto` for dumping fchk, which can also be placed in your ris class. Could also have an all-in-one function `get_nto_and_save` which wraps those two.

If you don't want to split them, making save_fch an option in your current function is also ok i guess, which at least allow people get the NTOs when they don't have mokit.

John-zzh:
It seems install from defaults channel also works,  with cupy==13.6.0 cutensor ==2.4.1.4


```
mamba create -n mkl -c defaults python=3.11
mamba activate mkl

mamba install -c defaults \
 cuda-version=12.8 \
 cupy=13 \
 cutensor=2 \
 numpy=2.2 \
 scipy \
 blas=*=*mkl
```

gpu4pyscf/tdscf/tests/test_ris.py
```diff
@@ -1,203 +1,231 @@
...
+        atom = '''
+        C         -4.89126        3.29770        0.00029
+        H         -5.28213        3.05494       -1.01161
+        O         -3.49307        3.28429       -0.00328
+        H         -5.28213        2.58374        0.75736
+        H         -5.23998        4.31540        0.27138
+        H         -3.22959        2.35981       -0.24953
+        '''
+        mol = gto.M(atom=atom, basis='def2-svp',
+                    # output = '/dev/null',  # Suppress excessive log output
```
sunqm: Add the `output='/dev/null'`. stdout.close() causes IO error

PR #640, opened 2026-01-23
Unrestricted KS vv10 gradient and hessian

henryw7:
Also support UKS (with or without vv10) gradient and hessian with grid response, and fix a bug with Becke weight derivatives.

### DIFF
--- a/gpu4pyscf/grad/tests/test_vv10_grid.py
+++ b/gpu4pyscf/grad/tests/test_vv10_grid.py
@@ -19,6 +19,7 @@ import pyscf
 from gpu4pyscf.dft import rks, uks
 from gpu4pyscf.grad.rks import _get_denlc
 from gpu4pyscf.lib.multi_gpu import num_devices
+from packaging.version import Version
 
 def setUpModule():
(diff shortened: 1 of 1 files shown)


gpu4pyscf/grad/tests/test_vv10_grid.py
```diff
@@ -197,6 +222,106 @@ def test_wb97xv_dense_grid_with_response(self):
...
+        # reference_gradient = numerical_denlc(mf, dm = None, denlc_only = False)
+        reference_gradient = np.array([
+            [ 0.066433938528121 , -0.04430744517947  , -0.0000001605826583],
+            [ 0.000320001092291 , -0.0102681219971146, -0.0000000675015599],
+            [-0.0625973459733586,  0.0470185344880747, -0.0000000547117907],
+            [-0.004156134636446 ,  0.0075570266488967, -0.000000072475359 ],
+        ])
+        test_gradient = grad_obj.kernel()
+
+        assert np.linalg.norm(test_gradient - reference_gradient) < 2e-6
```
sunqm, line 326: Add a test for `to_cpu()` then `to_gpu()` conversion, if the grid_response for vv10 is available in pyscf

PR #651, opened 2026-02-09
FSSH-TDDFT

sunqm:
A follow-up to PR #493.

* Refactor the implementation, remove duplicated functions.
* Move the fssh module to md module.
* API cleanups
  - Improve TDDFT scanner, using the X,Y amplitudes from previous step as initial guess.
  - Separate the FSSH-TDDFT into a subclass of FSSH base class
* Add tests and examples

### DIFF
--- a/gpu4pyscf/tdscf/rhf.py
+++ b/gpu4pyscf/tdscf/rhf.py
@@ -610,7 +610,10 @@ class TDA(TDBase):
         if x0 is None:
             if self.xy is None:
                 x0 = self.init_guess()
-            else: # Reuse the previous step for initial guess
+            else:
+                # Reuse the previous step for initial guess.
+                # Note, if the singlet attribute is altered from the previous
+                # run, reusing the initial guess may lead to convergence issues
                 x0 = self.xy
 
         if isinstance(x0, list):
--- a/gpu4pyscf/tdscf/tests/test_diffuse_orbital_tdscf.py
+++ b/gpu4pyscf/tdscf/tests/test_diffuse_orbital_tdscf.py
@@ -94,6 +94,7 @@ class KnownValues(unittest.TestCase):
         gobj.state = 1 # CIS_STATE_DERIV 3
         test_singlet_1_gradient = gobj.kernel()
 
+        td = mf.TDDFT().set(nstates = 3)
         td.singlet = False
         test_triplet_energies, _ = td.kernel()
         assert np.all(td.converged)


gpu4pyscf/nac/tests/test_tdrks_nac_scanner.py
```diff
@@ -15,10 +15,10 @@
 import unittest
 import numpy as np
 import cupy as cp
+import scipy.linalg
```
puzhichen: This is not used?

gpu4pyscf/nac/tdrhf.py
```diff
@@ -55,21 +55,18 @@ def match_and_reorder_mos(s12_ao, mo_coeff_b, mo_coeff, threshold=0.4):
...
         for idx in invalid_indices:
             print(f"Warning: reference coeff #{idx}'s best match is {final_chosen_overlaps[idx]:.4f} (below threshold {threshold})")
-    s_mo_new = mo_coeff_b.T @ s12_ao @ mo2_reordered
-    sign_array = cp.ones(s_mo_new.shape[-1])
-    for i in range(s_mo_new.shape[-1]):
-        if s_mo_new[i,i] < 0.0:
-            # mo2_reordered[:,i] *= -1
-            sign_array[i] = -1
+
+    sign_array = np.sign(cp.asnumpy(final_chosen_overlaps))
```
puzhichen: (`match_and_reorder_mos` is not used, so the code may leave unchanged.)

Should we consider the reordering of MO coefficients here? In the original version, `s_mo_new` was re-calculated to ensure the phase (sign) of the reordered orbitals remains consistent with the reference.  Since `final_chosen_overlaps` is derived from absolute values, using `np.sign` on it will always return 1, which might skip the necessary phase correction. Even if the new `_wfn_overlap` function no longer relies on  `mo2_reordered`, leaving this broken `sign_array` logic here might cause hidden bugs if this function is reused elsewhere.

gpu4pyscf/nac/tdrhf.py
```diff
@@ -325,9 +322,10 @@ def fvind(x):
         fvind,
         mo_energy,
         mo_occ,
-        wvo/(EJ-EI), # only one spin, negative in cphf
+        wvo,
         max_cycle=td_nac.cphf_max_cycle,
         tol=td_nac.cphf_conv_tol)[0] # eq.(80) in Ref. [1]
+    z1 /= EJ-EI # only one spin, negative in cphf
```
puzhichen, line 322: This modification is critical when calculating combinations of multiple states. I'm concerned about the numerical stability of the two approaches—I should verify which one is more robust in these scenarios.

gpu4pyscf/nac/tdrhf.py
```diff
@@ -722,39 +732,40 @@ def __call__(self, mol_or_geom, states=None, **kwargs):
...
-        mo2_reordered = cp.asarray(mo2_reordered)
-        mo_coeff0 = cp.asarray(mo_coeff0)
+        s = asarray(gto.intor_cross('int1e_ovlp', mol0, mol))
+        mo_coeff = cp.asarray(self.base._scf.mo_coeff)
 
         # for the first state
         if states[0] != 0: # excited state
-            sign = check_phase_modified(mol0, mo_coeff0, mo2_reordered, xi0, xi1, nocc, s)
-            self.sign *= np.sign(sign)
+            state_ovlp = _wfn_overlap(mo_coeff0, mo_coeff, xi0, xi1, s)
```
puzhichen, line 744: I noticed we are passing the un-reordered `mo_coeff` here instead of `mo2_reordered`. I assume this is intentional because the excitation amplitudes xi1 are in the basis of the original `mo_coeff`, and the Plasser algorithm (`_wfn_overlap`) naturally accounts for any orbital permutations and phase changes via the determinant of the raw overlap matrix. Could you confirm if my understanding is correct?

gpu4pyscf/md/distributions.py
```diff
@@ -0,0 +1,46 @@
...
+    Returns:
+        An (N, 3) array for velocities in atomic units (1 au = 21.877 A/fs)
+    '''
+    k_b = BOLTZMANN / HARTREE2J # = 8.314/2626e3
+    N = len(masses)
+    masses_amu = np.array(masses) * AMU2AU
+    std_dev = np.sqrt(k_b * temperature / masses_amu[:, np.newaxis])
+    velocities = np.random.normal(0, std_dev, (N, 3))
+    if force_temp:
+        total_momentum = np.sum(velocities * masses_amu[:, np.newaxis], axis=0)
```
puzhichen, line 39: Could we consider moving the center-of-mass momentum removal outside the `if force_temp` block to prevent the molecule from drifting? I understand our current code always uses `force_temp=True`, but this would make the function more robust if the default `False` is ever used in the future.

puzhichen, line 39: Could we consider moving the center-of-mass momentum removal outside the `if force_temp:` block to prevent the molecule from drifting? I understand our current code always uses `force_temp=True`, but this would make the function more robust if the default `False` is ever used in the future.

gpu4pyscf/md/fssh.py
```diff
@@ -0,0 +1,827 @@
...
+                hopping. DOI: 10.1063/1.2715585
+            [2] Nonadiabatic excited-state molecular dynamics: Treatment of
+                electronic decoherence. DOI: 10.1063/1.4809568
+
+        c_j = c_j * exp(-dt / tau_ji)
+        c_i = c_i * sqrt((1 - sum_j(j!=i) |c_j|**2) / |c_i|**2)
+        tau_ji = ħ / |E_jj - E_ii| * (1 + a / E_kin)
+        """
+
+        E_kin = (0.5 * self.mass[:,None] * np.sum(velocity ** 2)).sum()
```
puzhichen: Is this  bug in the kinetic energy calculation here? Because `np.sum(velocity ** 2)` is evaluated first, it collapses the entire velocity array into a single scalar before multiplying by the mass array. This does not represent the correct physical kinetic energy ($\frac{1}{2} \sum m_i v_i^2$).

gpu4pyscf/md/fssh.py
```diff
@@ -0,0 +1,827 @@
...
+        κ_ij = sqrt(max(0, d²(E_j - E_i)/dt² / (E_j - E_i)))
+
+        where the second derivative is approximated using three-point finite differences:
+        d²(ΔE)/dt² ≈ (ΔE_t - 2*ΔE_p + ΔE_pp) / dt²
+
+        Args:
+            energy_t (np.ndarray): Time-dependent energy array (Nstates,)
+            energy_p (np.ndarray): Time-dependent energy array (Nstates,)
+            energy_pp (np.ndarray): Time-dependent energy array (Nstates,)
+
```
puzhichen, line 167: Could we make the docstrings for `energy_t`, `energy_p`, and `energy_pp` a bit more specific? Since they currently share the exact same description, it might be very helpful for future users if we explicitly state their time steps (e.g., current step, previous step, and two steps ago).

gpu4pyscf/md/fssh.py
```diff
@@ -0,0 +1,827 @@
...
+            hop_index = self.evaluate_hopping(coefficient, nact, cur_state)
+
+            # 6. adjust nuclear velocity
+            cur_idx = self.states.index(cur_state)
+            if hop_index != -1 and hop_index != cur_idx:
+
+                # Calculate d_vec for velocity rescaling
+                if self.coupling_method in ('nac', 'direct'):
+                    d_vec = pes.nacv[cur_idx, hop_index]
+                elif self.coupling_method == ('ktdc', 'curvature'):
```
puzhichen, line 720: `self.coupling_method` is a string.

gpu4pyscf/md/fssh_tddft.py
```diff
@@ -0,0 +1,157 @@
...
+            if abs(state_ovlp) < 0.3:
+                states_reorder.append(i)
+            else:
+                self._sign[i] *= np.sign(state_ovlp)
+
+        states = self.states
+        nstates = len(states)
+        # Indices for nonadiabatic coupling calculations. By default,
+        # all pairs (i,j) where i < j within self.states are evaluated.
+        nac_idx = [(i,j) for i in range(nstates-1) for j in range(i+1, nstates)]
```
puzhichen: I think we can exclude the ground state from this calculation. Should we hardcode the logic here to only compute the NACVs between excited states?

gpu4pyscf/md/fssh_tddft.py
```diff
@@ -0,0 +1,157 @@
...
+
+        if with_nacv:
+            # Save the previous states to determine the phase of NACVs
+            mo_coeff0 = cp.asarray(mf.mo_coeff)
+            nmo = mo_coeff0.shape[1]
+            nocc = int((mf.mo_occ > 0).sum())
+            nvir = nmo - nocc
+            if isinstance(td_scanner, RisBase):
+                xs0 = td_scanner.xy[0]
+                xs0 = [xs0[i-1].reshape(nocc, nvir) for i in self.states]
```
puzhichen: If ground state index 0 is in `self.states`, will this part be problematic?

gpu4pyscf/md/fssh_tddft.py
```diff
@@ -0,0 +1,157 @@
...
+        else:
+            xs1 = [td_scanner.xy[i-1][0] for i in self.states]
+
+        s = cp.asarray(gto.intor_cross('int1e_ovlp', mol0, mol))
+        s_mo_ground = mo_coeff0[:, :nocc].T.dot(s).dot(mo_coeff[:, :nocc])
+        self._sign[0] *= np.sign(cp.linalg.det(s_mo_ground).get())
+
+        states_reorder = []
+        for i in self.states:
+            state_ovlp = _wfn_overlap(mo_coeff0, mo_coeff, xs0[i-1], xs1[i-1], s)
```
puzhichen, line 117: Can self.states be non-contiguous? The lengths of `xs0` and `xs1` are `len(self.states)`. Indexin with` i-1` here may be out of bounds.

examples/md/01-fssh_td_ris.py
```diff
@@ -0,0 +1,65 @@
...
+
+# Initial velocities from a Maxwell–Boltzmann distribution at 300 K.
+# Initial positions and velocities can also be generated using the Wigner
+# sampling method (see 02-wigner_sampling.py).
+masses = mol.atom_mass_list()
+v = maxwell_boltzmann_velocities(masses, temperature=300, force_temp=True)
+
+# Run a FSSH simulation including only the first and second excited
+# electronic states. This restricts surface hopping among these states.
+fssh = FSSH_TDDFT(td, [1, 2])
```
puzhichen, line 51: Should we hardcode to exclude the ground state?

gpu4pyscf/md/wigner_sampling.py
```diff
@@ -0,0 +1,99 @@
...
+        samples:
+            Number of initial samples to generate
+
+    Returns:
+        A list of (positions, velocities)
+    '''
+    if seed is not None:
+        np.random.seed(seed)
+    valid = []
+    temperature = 300
```
puzhichen: The input parameter `temp` does not work, since `temperature = 300` is hardcoded.

gpu4pyscf/md/wigner_sampling.py
```diff
@@ -0,0 +1,99 @@
...
+    ex = mu / temp * wavenumber2Kelvin  # vibrational temperature: ex=h*c*mu/(kb*T), 0.69503 convert cm-1 to K
+    pop = []
+
+    while np.sum(pop) < max_pop and len(pop) < max_nlevel:
+        _pop = float(np.exp(-1 * ex * len(pop)) * (1 - np.exp(-1 * ex)))
+        pop.append(_pop)
+    pop.append(1 - np.sum(pop))
+
+    if seed is not None:
+        np.random.seed(seed)
```
puzhichen: If the `seed` is set here, each `wignerfunc` call 
```
q = np.random.uniform(0, 1) * 10.0 - 5.0
p = np.random.uniform(0, 1) * 10.0 - 5.0
```
wii give the same result.

examples/md/01-fssh_td_ris.py
felixmiaomiao: Why should we exclude the ground state? In some cases, the results of studies related to GS are acceptable.

examples/md/01-fssh_td_ris.py
```diff
@@ -0,0 +1,65 @@
...
+# Set the initial electronic state to the second excited state.
+fssh.cur_state = 2
+fssh.nsteps = 50 # Number of time steps to propagate.
+fssh.timestep_fs = 1.0 # fs
```
felixmiaomiao: I think 1.0 fs is too large. 0.5 fs will an acceptable timestep

gpu4pyscf/md/fssh_tddft.py
felixmiaomiao: In program, `i=0` will result in `xs0[-1]`, which will run without raising error, however correctness need further checking.

gpu4pyscf/md/fssh.py
felixmiaomiao: This is indeed a bug here. It should be `E_kin = (0.5 * self.mass[:,None] *velocity ** 2).sum()`

gpu4pyscf/md/fssh.py
```diff
@@ -0,0 +1,827 @@
...
+                 f"dt={self.timestep_fs:.3f} fs, total steps: {self.nsteps}\n"
+                 f"coupling_method={self.coupling_method}\n"
+                 f"decoherence={self.decoherence}, alpha={self.alpha}\n"
+                 f"Trajectory will be saved to {self.filename}\n")
+        log.info("Starting FSSH trajectory simulation")
+
+        if self.seed is not None:
+            np.random.seed(self.seed)
+
+        if position is None:
```
felixmiaomiao, line 632: Is it necessary to allow external input for positions?

gpu4pyscf/md/fssh.py
```diff
@@ -0,0 +1,827 @@
...
+                hop_allowed, velocity = self.rescale_velocity(
+                    hop_index, cur_state, pes.energy, velocity, d_vec)
+
+                if hop_allowed:
+                    cur_state, old_state = self.states[hop_index], cur_state
+
+                    log.info(f"Hop: {old_state} → {cur_state} at step {step}")
+
+                else:
+                    ke = (0.5 * self.mass[:,None] * velocity ** 2).sum()
```
felixmiaomiao: This is actually the insufficient kinetic energy along the `d_vec` direction.

gpu4pyscf/md/fssh.py
```diff
@@ -0,0 +1,827 @@
...
+                self.callback(locals())
+
+            current_idx = self.states.index(self.cur_state)
+            current_energy = pes.energy[current_idx]
+            populations = np.abs(self.coefficient)**2
+
+            # Format output
+            total_time = step * self.timestep_fs
+            log.info(f"Step {step:4d}: Time {total_time:8.3f} fs, State {cur_state:2d}, "
+                     f"Energy {current_energy:12.8f} Ha, Populations: {populations}")
```
felixmiaomiao, line 777: Will it be better if Energy display all selected states.

examples/md/01-fssh_td_ris.py
puzhichen: > Why should we exclude the ground state? In some cases, the results of studies related to GS are acceptable.
I agree that GS results are valuable, but my concern is that our current computational method might not correctly handle the coupling between the ground state and excited states. To ensure the reliability for now, I thought we could temporarily exclude the GS until we refine the coupling calculations by spin-flip calculations. What do you think?

gpu4pyscf/md/fssh.py
```diff
@@ -0,0 +1,827 @@
...
+            step (int): Current simulation step
+            position (np.ndarray): Nuclear coordinates in Bohr
+            velocity (np.ndarray): Nuclear velocities in atomic units
+            pes (PES): Electronic structure energy, force, and nacv
+            coeffs (np.ndarray): Quantum coefficients
+            cur_state (int): Current active electronic state
+        """
+        mode = 'w' if step == 0 else 'a'
+
+        with h5py.File(self.filename, mode) as f:
```
felixmiaomiao: This is a time-consuming operation. Can we open the file at the beginning and close it at the end of the whole simulation? New data can be written into the file using functions like 
`h5file = h5py.File(self.filename, mode) 
h5file.flush()`

gpu4pyscf/nac/tdrhf.py
sunqm: Fixed

gpu4pyscf/nac/tdrhf.py
sunqm: Reordering MO is not necessary since the sign is determined by the determinant evaluation in the _wfn_overlap function

gpu4pyscf/md/fssh_tddft.py
sunqm: Ground state (i=0) is supported now

examples/md/01-fssh_td_ris.py
sunqm: GS is supported in the interface, although the NACVs between excited states and GS might not be accurate

gpu4pyscf/md/distributions.py
sunqm: Adjusting the center-of-mass momentum will break the temperature. This parameter in the API follows the convention of the corresponding functions in ASE.

gpu4pyscf/md/wigner_sampling.py
sunqm: seed has been removed

gpu4pyscf/md/fssh.py
sunqm: This input is not necessary for molecules. However, it is kept to make the base class API more general for other use cases, even if it is redundant in this specific case.

gpu4pyscf/md/fssh.py
sunqm: printed in logger

gpu4pyscf/md/fssh.py
sunqm: supported

gpu4pyscf/md/wigner_sampling.py
sunqm: fixed

gpu4pyscf/md/fssh.py
sunqm: Fixed

PR #657, opened 2026-02-17
upgrade pyscf/dispersion and make gpu4pyscf self-contain all dispersion codes

JiashuLiang:


gpu4pyscf/dispersion/__init__.py
```diff
@@ -0,0 +1 @@
+__version__ = '1.5.0'
```
sunqm: This version Id is likely not necessary.

If the version Id is designed to indicate the corresponding pyscf-dispersion release, consider to import `__version__`  from pyscf-dispersion

sunqm:
Could you please provide an example to demonstrate the upgraded "disp" convention?

gpu4pyscf/dispersion/__init__.py
JiashuLiang: Okay, fixed

PR #682, opened 2026-03-09
Fix the wrong sign for external field

henryw7:


gpu4pyscf/qmmm/external_field.py
```diff
@@ -90,7 +90,7 @@ def get_hcore(self, mol=None):
         if self.electric_field is not None:
             with mol.with_common_orig(self.origin):
                 dipole_integral = cp.asarray(mol.intor('int1e_r'))
-            h1e -= cp.einsum('d,dij->ij', self.electric_field, dipole_integral)
+            h1e += cp.einsum('d,dij->ij', self.electric_field, dipole_integral)
```
sunqm, line 93: The dipole of the molecule = `-dipole_integral`, the interaction of electric field and dipole should be -E*d

henryw7: The potential energy for any charge q in a uniform electric field E is U = -qEr. An electron has charge -1 in au, so U = Er, which has no negative sign.

PR #710, opened 2026-03-26
cuEST interface against nvidia-cuest-cu13==0.1.0

henryw7:


gpu4pyscf/lib/cuest_wrapper.py
```diff
@@ -0,0 +1,4942 @@
...
+
+        self._locked_keys = ['atom_grid', 'level', 'prune', 'radi_method', 'becke_scheme', 'atomic_radii', 'radii_adjust']
+
+        self.verbose = log.verbose
+        self.coords = np.array([np.nan])
+        self.weights = np.array([np.nan])
+
+        # Intentionally not calling super().__init__()
+
+    def __setattr__(self, key, val):
```
sunqm, line 3788: Maybe use `@property` for these attributes instead.

gpu4pyscf/lib/cuest_wrapper.py
```diff
@@ -0,0 +1,4942 @@
...
+            logger.info(self, 'atomic radii adjust function: %s', self.radii_adjust)
+            logger.debug2(self, 'atomic_radii : %s', self.atomic_radii)
+        if self.atom_grid:
+            logger.info(self, 'User specified grid scheme %s', str(self.atom_grid))
+        return self
+
+    def _do_nothing(self, *args, **kwargs):
+        return self
+    build = _do_nothing
+    reset = _do_nothing
```
sunqm, line 3808: Letting `reset`, `to_cpu` and `to_gpu` fail might be better

gpu4pyscf/lib/cuest_wrapper.py
```diff
@@ -0,0 +1,4942 @@
...
+
+        # Intentionally not calling super().__init__()
+
+    def __setattr__(self, key, val):
+        if key in self._locked_keys:
+            raise RuntimeError("The numint setup should not be changed after CuESTWrapper is applied.")
+        super().__setattr__(key, val)
+
+    def nr_rks(self, mol, grids, xc_code, dms, relativity=0, hermi=1, max_memory=0, verbose=None):
+        assert relativity == 0, "Relativistic KS is not supported by cuEST yet"
```
sunqm, line 3843: This parameter can be ignored. This legacy parameter is actually not used anywhere in pyscf or gpu4pyscf

gpu4pyscf/lib/cuest_wrapper.py
```diff
@@ -0,0 +1,4942 @@
...
+
+    if isinstance(method, CuESTWrapper):
+        method.reset()
+        return method
+
+    cls = CuESTWrapper
+
+    return lib.set_class(cls(method), (cls, method.__class__))
+
+def apply_cuest_gradient_wrapper(grad_method):
```
sunqm, line 4901: Should the `.base` of `grad_method` be also converted to a cuest instance?

gpu4pyscf/lib/cuest_wrapper.py
```diff
@@ -0,0 +1,4942 @@
...
+
+    def to_gpu(self):
+        return self
+
+    def to_cpu(self):
+        raise NotImplementedError("CuESTWrapper does not support to_cpu() method. Please reconstruct your object.")
+
+def apply_cuest_wrapper(method):
+    import gpu4pyscf
+    assert (isinstance(method, gpu4pyscf.scf.hf.SCF))
```
sunqm, line 4891: This can be a factory function, automatically calling the corresponding wrapper if gradient object or other types of objects are provided.

gpu4pyscf/lib/cuest_wrapper.py
henryw7: Here https://github.com/bytedance/gpu4pyscf/blob/647dc63bb4f53a801fd5647777d87b3728feb7ad/gpu4pyscf/lib/cuest_wrapper.py#L4900:L4901 I've checked that the base object has to be a cuest instance.

gpu4pyscf/lib/cuest_wrapper.py
henryw7: I think it's fine to have this check, unless we remove this argument from all pyscf/gpu4pyscf apis.

gpu4pyscf/lib/cuest_wrapper.py
henryw7: I don't want it to be too general-purposed, since basically nothing other than mean-field and mean-field gradient is supported. And I want user to get gradient via mf.Gradients() method, not by passing in a non-cuest mean-field. We can change the behavior later, when cuest supports more things.

gpu4pyscf/lib/cuest_wrapper.py
henryw7: I've disabled all to_cpu and to_gpu functions of the underlying Grids/Numint/PCM classes. For reset in Grid, grids.reset() will be called in normal code path, and I didn't store anything (mol/actual grid geometry) in the CuESTExtractedGrid object, so it's fine to do nothing. For reset in Numint and PCM, they are disabled.

gpu4pyscf/lib/cuest_wrapper.py
henryw7: I think I'm writing more code to achieve the same effect with `@property`?

PR #719, opened 2026-04-01
Support explicit dispersion and nonlocal correlation specification in EDA interface

henryw7:
- Fix incorrect dispersion for EDA
- Fix cupy interface change in cupy 14.0.0


gpu4pyscf/properties/eda.py
```diff
@@ -30,6 +30,7 @@
 from gpu4pyscf.lib import logger
 import time
 import warnings
+from packaging.version import Version
```
sunqm, line 33: In setup.py, add "packaging" into the install dependency

PR #778, opened 2026-05-29
Handle wb97x-d4

sunqm:
Corresponding to https://github.com/pyscf/pyscf/pull/3224

### DIFF
--- a/gpu4pyscf/scf/dispersion.py
+++ b/gpu4pyscf/scf/dispersion.py
@@ -23,7 +23,7 @@ dispersion correction for HF and DFT
 import warnings
 from functools import lru_cache
 from pyscf.lib import logger
-from gpu4pyscf import scf
+from pyscf import scf
 from pyscf import __config__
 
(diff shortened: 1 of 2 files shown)


gpu4pyscf/scf/dispersion.py
```diff
@@ -72,14 +80,40 @@ def parse_dft(xc_code):
...
+
+DFTD4_RECOMMENDATIONS = True
+
+in ~/pyscf_conf.py`, or at runtime via
+
+pyscf.scf.dispersion.DFTD4_RECOMMENDATIONS = True
+
+The default behavior will change to the DFT-D4 recommended convention in PySCF v2.16.
+''', FutureWarning, stacklevel=2)
+        return 'wb97x', False, 'd4:wb97x'
```
JiashuLiang, line 112: 'd4:wb97x'  refers to the D4 parameter of wb97x-v in new dftd4 version. To reproduce old behavior, we need to use 'd4:wb97x-2008'

JiashuLiang: I create a PR 
https://github.com/pyscf/gpu4pyscf/pull/812

PR #791, opened 2026-06-16
Minor reorg of gout_pattern bug

abagusetty:
1. Make constant-memory vars const
2. multiple kernel launch are wrapped under lambdas

### DIFF
--- a/gpu4pyscf/lib/gvhf-rys/rys_contract_jk.cu
+++ b/gpu4pyscf/lib/gvhf-rys/rys_contract_jk.cu
@@ -17,6 +17,7 @@
 #include <stdio.h>
 #include <stdint.h>
 #include <stdlib.h>
+#include <type_traits>
 #include <cuda_runtime.h>
 
 #include "gint/cuda_alloc.cuh"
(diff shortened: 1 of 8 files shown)


gpu4pyscf/lib/gvhf-rys/rys_contract_jk.cu
```diff
@@ -586,40 +588,31 @@ int RYS_build_jk(double *vj, double *vk, double *dm, int n_dm, int nao,
...
+        auto launch = [&](auto offset, int tile_chunk) {
+            constexpr int OFFSET = decltype(offset)::value;
+            int tdims[2];
+            size_t buflen = threads_scheme_for_jk(tdims, bounds, shm_size, tile_chunk);
+            dim3 threads;
+            threads.x = tdims[0];
+            threads.y = tdims[1];
             int reserved_shm_size = (buflen - cart_idx_size*4)/8;
-            rys_jk_kernel<<<workers, threads, buflen>>>(
+            rys_jk_kernel<OFFSET><<<workers, threads, buflen>>>(
```
sunqm, line 606: Why introduce a template for this function?

The compile-time constant offset seems not providing any advantage over passing the address to the function.

sunqm (approved):

gpu4pyscf/lib/gvhf-rys/rys_contract_jk.cu
abagusetty: Thanks @sunqm for taking time to review the PR. The approach behind using templated offset is due to setting "constant-memory variables pointer arithmetic" limitation: `p_gxyz_offset+256` & `p_gxyz_offset+512` in the kernel launch for SYCL. 

CUDA allows the pointer arithmetic of a device pointer in the args for the kernel launch.
SYCL unfortunately has a limitation. Since `p_gxyz_offset` is a constant memory "device" pointer, pointer arithmetic at the kernel launch site is still treated as a host-side pointer and any pointer arithmetic is not valid. Hence I had to pass OFFSET as a template arg (not much of a benefit, I agree) and do this `const GXYZOffset *gxyz_offsets = p_gxyz_offsets + OFFSET;` inside the kernel definition.

Mostly looking for a common solution that works for CUDA and SYCL (in future for HIP backend). Please let me know if you prefer an other approach, happy to take a crack.

PR #804, opened 2026-06-25
Fix the OOM problem introduced by low ke_cutoff

henryw7:
When `ke_cutoff` is low, all shells with high ke will be placed into the same bucket (task), and that particular bucket requires a large amount of GPU memory for the Fock matrix in diff_image space.


sunqm:
Please add a test for the limited memory case
```
with lib.temporary_env(multgrid_v2, get_avail_mem=(lambda **kw: 1000000)):
    ...
```

sunqm (approved):

PR #812, opened 2026-07-05
Fix wb97x D4 method parsing

JiashuLiang:
Fix the bug in PR #778

### DIFF
--- a/gpu4pyscf/scf/dispersion.py
+++ b/gpu4pyscf/scf/dispersion.py
@@ -109,13 +109,17 @@ pyscf.scf.dispersion.DFTD4_RECOMMENDATIONS = True
 
 The default behavior will change to the DFT-D4 recommended convention in PySCF v2.16.
 ''', FutureWarning, stacklevel=2)
-        return 'wb97x', False, 'd4:wb97x'
+        return 'wb97x', False, 'd4:wb97x-2008'
 
     if method_lower in _white_list:
         return _white_list[method_lower]
 
-    if '-d3' in method_lower or '-d4' in method_lower:
-        xc, disp = method_lower.split('-')
+    if '-d3' in method_lower:
+        xc, _, disp_suffix = method_lower.partition('-d3')
+        disp = f'd3{disp_suffix}'
+    elif '-d4' in method_lower:
+        xc, _, disp_suffix = method_lower.partition('-d4')
+        disp = f'd4{disp_suffix}'
     else:
         xc, disp = method_lower, None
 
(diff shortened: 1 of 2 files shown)


sunqm (approved):

gpu4pyscf/scf/dispersion.py
```diff
@@ -109,13 +109,17 @@ def parse_dft(xc_code):
...
+        return 'wb97x', False, 'd4:wb97x-2008'
 
     if method_lower in _white_list:
         return _white_list[method_lower]
 
-    if '-d3' in method_lower or '-d4' in method_lower:
-        xc, disp = method_lower.split('-')
+    if '-d3' in method_lower:
+        xc, _, disp_suffix = method_lower.partition('-d3')
+        disp = f'd3{disp_suffix}'
```
sunqm, line 119: should this be `f'd3:{disp_suffix}'`

gpu4pyscf/scf/dispersion.py
```diff
@@ -109,13 +109,17 @@ def parse_dft(xc_code):
...
+    elif '-d4' in method_lower:
```
sunqm, line 120: cannot parse `wb97x-d4s`

gpu4pyscf/scf/dispersion.py
JiashuLiang: No, as
 'wb97x-d3:pbe'.partition('-d3')
  # ('wb97x', '-d3', ':pbe')
  # -> 'd3:pbe'

gpu4pyscf/scf/dispersion.py
JiashuLiang: Now it is parsed as 'wb97x', False, 'd4:wb97x'
see gpu4pyscf/scf/tests/test_dispersion_logic.py:70
        self.assertEqual(dispersion.parse_disp('wb97x-d4s'), ('wb97x', 'd4s', True))

Do you mean that we should add it to _white_list so that it is parsed as
    'wb97x_v', False, 'd4s:wb97x'
 Or 
    'wb97x', False, 'd4s:wb97x-2008'

sunqm: Does it support `parse_disp('wb97x-d4s')`, `parse_disp('wb97x-3c')` ?

JiashuLiang: Yes 
```
  parse_disp('wb97x-d4s')
  # ('wb97x', 'd4s', True) # This is different from neither pyscf tradition, nor academical tradition. We may need to change.

  parse_disp('wb97x-3c')
  # ('wb97x-3c', 'd4', True)
```

PR #823, opened 2026-07-16
Make the add_sparse() function more tolerant to wrong-order input

henryw7:
When using cupy (not cutensor) for contract() function, the returned matrix may not be in C-order.
It's ok to reshape b input of add_sparse(), since it is a unmodified input.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,10 @@ There shouldn't be cupy or cutensor compilation during pip install process. If y
 ```
 <repo_path>/gpu4pyscf/lib/cutensor.py:<line_number>: UserWarning: using cupy as the tensor contraction engine.
 ```
+The following command may help when using `cupy-cuda12x==14.1.1` and `cutensor-cu12==2.7.0`:
+```sh
+export LD_LIBRARY_PATH="$(python -c 'import cutensor; print(cutensor.__path__._path[0])')/lib:${LD_LIBRARY_PATH}"
+```
 
 Features
 --------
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -52,6 +52,10 @@ There shouldn't be cupy or cutensor compilation during pip install process. If y
 ```
 <repo_path>/gpu4pyscf/lib/cutensor.py:<line_number>: UserWarning: using cupy as the tensor contraction engine.
 ```
+The following command may help when using `cupy-cuda12x==14.1.1` and `cutensor-cu12==2.7.0`:
+```sh
+export LD_LIBRARY_PATH="$(python -c 'import cutensor; print(cutensor.__path__._path[0])')/lib:${LD_LIBRARY_PATH}"
```
sunqm: Setting LD_LIBRARY_PATH is misleading and should be removed.

cupy has its own logic for locating cutensor library. It does not rely on LD_LIBRARY_PATH

henryw7: Yes agreed, only if cupy logic works. For the recent cupy version (14.1.*) it has trouble finding the cutensor library, and setting `LD_LIBRARY_PATH` fixes it.

sunqm: https://github.com/cupy/cupy/blob/9771347977f79ca08ebf0fb7fb6250268eaa35f5/cupy/_environment.py#L441
here is how cupy searches for cutensor. If cutensor is installed from wheel, it is the prior one. This behavior is not changed in cupy 14.*. Configuring `LD_LIBRARY_PATH` is unlikely the right solution.

Please open an issue describing this problem and the local packages you have installed. We can explore in more details there

sunqm (approved):

PR #828, opened 2026-07-22
Fix stale auxmol after `DF.reset()` with in-place geometry change

### LINKED ISSUE #827: `DF.reset()` keeps a stale `auxmol` after in-place geometry change — silently wrong SCF energies (diverges from PySCF CPU behavior)
### Summary

After changing the geometry of `mf.mol` in place (`mol.set_geom_(...)`) and
calling `mf.reset()`, a density-fitted calculation in gpu4pyscf rebuilds
`_cderi` **with the auxiliary basis still sitting at the old coordinates**.
The SCF then converges normally and returns a silently wrong energy — for a
1e-3 Bohr displacement of one atom in water/def2-SVP we get an error of
**~2e-2 Ha**, with no warning or error raised.

The same sequence is handled correctly by CPU PySCF, so code that follows the
documented scanner/reset pattern gives correct results on CPU and corrupted
results on GPU. We hit this in production: our finite-difference Hessians
(displace → `reset()` → `kernel()` loop) were entirely corrupted, producing
non-physical frequencies in the 10^4–10^5 cm⁻¹ range before we tracked it down.

### Minimal reproduction

```python
from pyscf import gto, dft

WATER = "O 0 0 0; H 0.9572 0 0; H -0.2400 0.9266 0"

def make_mol(displaced=False):
    mol = gto.M(atom=WATER, basis="def2-svp", unit="Angstrom", verbose=0)
    if displaced:
        c = mol.atom_coords(unit="Bohr").copy()
        c[0, 0] += 1e-3
        mol.set_geom_(c, unit="Bohr")
        mol.build(dump_input=False)
    return mol

# Reference: fresh mf at the displaced geometry
e_fresh = dft.RKS(make_mol(True), xc="wb97x").density_fit().to_gpu().kernel()

# Reuse: kernel at equilibrium, displace in place, reset, kernel again
mf = dft.RKS(make_mol(), xc="wb97x").density_fit().to_gpu()
mf.kernel()
c = mf.mol.atom_coords(unit="Bohr").copy()
c[0, 0] += 1e-3
mf.mol.set_geom_(c, unit="Bohr")
mf.mol.build(dump_input=False)
mf.reset()
e_reused = mf.kernel()

print(f"fresh  = {float(e_fresh):.10f}")
print(f"reused = {float(e_reused):.10f}")
print(f"|diff| = {abs(float(e_reused) - float(e_fresh)):.2e} Ha")
# stale auxiliary basis: max deviation equals the displacement
import numpy as np
aux = mf.with_df.auxmol.atom_coords(unit="Bohr")
print(f"auxmol max deviation from current mol geometry: "
      f"{np.abs(aux - mf.mol.atom_coords(unit='Bohr')).max():.2e} Bohr")
```

Observed (gpu4pyscf 1.7.4, cupy 14.1.1, pyscf 2.13.1, RTX 3070, CUDA 12.4/WSL2):

```
fresh  = -76.3496849152
reused = -76.3700807832
|diff| = 2.03e-02 Ha
auxmol max deviation from current mol geometry: 1.00e-03 Bohr
```

Expected: `|diff|` at SCF-convergence level (the same script with
`.density_fit()` removed, or run on CPU PySCF, gives ~1e-11 Ha).

Controls we ran (same displace+reset+kernel sequence):

| variant            | result |
|--------------------|--------|
| GPU DF + SMD       | wrong (2.04e-2 Ha) |
| GPU DF only        | wrong (2.03e-2 Ha) |
| GPU SMD only (no DF) | correct |
| GPU bare RKS       | correct |
| CPU DF (+/- SMD)   | correct |

So the issue is specific to the density-fitting path on GPU; solvent models
are unaffected on their own.

### Root cause

CPU PySCF `df.DF.build` regenerates the auxiliary Mole **unconditionally**
from the current `self.mol` on every build
([pyscf/df/df.py](https://github.com/pyscf/pyscf/blob/master/pyscf/df/df.py),
`auxmol = self.auxmol = addons.make_auxmol(self.mol, self.auxbasis)`),
so a stale `auxmol` can never survive a rebuild.

gpu4pyscf `df.DF.build` only regenerates it when it is missing:

```python
if auxmol is None:
    self.auxmol = auxmol = addons.make_auxmol(mol, self.auxbasis)
```

and `DF.reset(mol=None)` only clears `auxmol` when a new `mol` object is
passed:

```python
def reset(self, mol=None):
    if mol is not None:
        self.mol = mol
        self.auxmol = None
        ...
    self._cderi = None
    ...
```

`mf.reset()` (no argument) therefore keeps the old-geometry `auxmol` (and on
current master also `intopt`/`j_engine`), while `_cderi` is rebuilt against
it — producing wrong 3c2e integrals for the new geometry with no diagnostic.

### Suggested fix

Either
- clear `self.auxmol` (and `intopt`/`j_engine`) unconditionally in
  `DF.reset()`, or
- rebuild `auxmol` from `self.mol` unconditionally in `DF.build()`, matching
  CPU PySCF semantics.

If keeping the cache is intentional for scanner performance, a geometry
consistency check between `mol` and `auxmol` at build time would at least
turn silent corruption into a loud error.

### Environment

Minimal reproduction above was run on:

- gpu4pyscf 1.7.4 (code path unchanged on current master as of 2026-07-23)
- pyscf 2.13.1, cupy 14.1.1
- NVIDIA GeForce RTX 3070, CUDA 12.4, WSL2 (Linux 6.6)

The original production corruption (finite-difference Hessians) was also
observed on RTX 4070 Ti and A100 SXM systems with CUDA 13.x, so the issue is
not specific to a particular GPU architecture or CUDA major version.

### Possibly related

#552 reported numerical vs analytical Hessian disagreement with DF on GPU.
If that workflow reused a live mf across displacements (in-place
`set_geom_` + `reset()`), the stale-`auxmol` behavior described here could
explain it.

Isitea:
Fixes #827.

### What happens

`mol.set_geom_()` (in place) + `mf.reset()` + `mf.kernel()` — the standard
finite-difference / reuse pattern — silently returns wrong SCF energies with
density fitting on GPU (~2e-2 Ha for a 1e-3 Bohr single-atom displacement in
water/def2-SVP), while the same sequence is correct on CPU PySCF. See #827
for the minimal reproduction and isolation table.

### Root cause (two layers)

1. `DF.reset(mol=None)` only discards `auxmol`/`intopt`/`j_engine` when a new
   `mol` object is passed. Calling `mf.reset()` without arguments keeps the
   auxiliary basis at the previous geometry while `_cderi` is rebuilt against
   it.
2. Even when `auxmol` is rebuilt, `make_auxmol` re-parses `mol._atom` (the
   input record). After `to_gpu()`, `mf.mol` and `mf.with_df.mol` are distinct
   objects sharing `_env`, so an in-place `mf.mol.set_geom_()` updates the
   runtime coordinates (`_env`) but not `with_df.mol._atom` — the rebuilt
   `auxmol` still sits at the old coordinates. (The AO side is unaffected:
   `SortedMole.from_mol` works from `mol._env`.)

### Fix

- `DF.reset()`: discard the geometry-dependent caches (`auxmol`, `intopt`,
  `j_engine`) unconditionally, matching CPU PySCF semantics where `DF.build`
  regenerates the auxiliary Mole on every build.
- `DF.build()`: after `make_auxmol`, sync the auxiliary basis centers to
  `mol.atom_coords()` so the auxiliary basis always follows the runtime
  coordinates. This is a no-op in the normal case and costs one `set_geom_`
  per build.

### Validation

Applying the equivalent patch to the 1.7.4 runtime (RTX 3070, CUDA 12.4) and
running the displace + reset + kernel sequence from #827:

```
before: |e_reused - e_fresh| = 2.03e-02 Ha   (silent corruption)
after:  |e_reused - e_fresh| = 4.30e-11 Ha   (SCF-convergence level)
```

### Tests

`gpu4pyscf/df/tests/test_df_reset.py` adds three regression tests:

- reuse energy after in-place displacement + reset matches a freshly built mf
  at the same geometry (< 1e-8 Ha),
- `reset()` clears `auxmol`/`intopt`,
- the rebuilt `auxmol` sits at the runtime coordinates of `mf.mol`.

### Notes for reviewers

If keeping `auxmol` across `reset()` is intentional for scanner performance,
an alternative would be a geometry-consistency check in `build()` that raises
instead of silently reusing stale coordinates — happy to rework in that
direction if preferred.

### DIFF
--- a/gpu4pyscf/df/df.py
+++ b/gpu4pyscf/df/df.py
@@ -224,14 +224,27 @@ class DF(lib.StreamObject):
         '''Reset mol and clean up relevant attributes for scanner mode'''
         if mol is not None:
             self.mol = mol
-        # auxmol, intopt and j_engine depend on the geometry of self.mol, so
-        # they must be discarded even when reset() is called without a new mol
-        # (e.g. after an in-place mol.set_geom_()). Keeping them rebuilds
-        # _cderi against stale auxiliary-basis coordinates and yields silently
-        # wrong SCF energies (issue #827).
-        self.auxmol = None
-        self.intopt = None
-        self.j_engine = None
+            self.auxmol = None
+            self.intopt = None
+            self.j_engine = None
+        elif self.auxmol is not None and self.mol is not None:
+            # An in-place mol.set_geom_() leaves auxmol/intopt/j_engine at the
+            # old coordinates; rebuilding _cderi against them yields silently
+            # wrong SCF energies (issue #827). Re-anchor auxmol to the runtime
+            # coordinates *in place* (several callers hold a reference to
+            # with_df.auxmol across reset() as a memory-release idiom, so the
+            # object must stay valid), and drop the geometry-bound integral
+            # engines only when the geometry actually moved.
+            if self.auxmol.natm != self.mol.natm:
+                self.auxmol = None
+                self.intopt = None
+                self.j_engine = None
+            else:
+                mol_coords = self.mol.atom_coords()
+                if abs(self.auxmol.atom_coords() - mol_coords).max() > 1e-10:
+                    self.auxmol.set_geom_(mol_coords, unit='Bohr')
+                    self.intopt = None
+                    self.j_engine = None
         self._cderi = None
         self._cderi_idx = None
         self._cd_j2c = None
--- a/gpu4pyscf/df/df.py
+++ b/gpu4pyscf/df/df.py
@@ -95,6 +95,11 @@ class DF(lib.StreamObject):
         t0 = log.init_timer()
         if auxmol is None:
             self.auxmol = auxmol = addons.make_auxmol(mol, self.auxbasis)
+            # make_auxmol re-parses mol._atom (the input record), which can be
+            # stale when the geometry was updated in place through another Mole
+            # sharing mol._env (e.g. mf.mol after to_gpu()). Sync the auxiliary
+            # basis centers to the runtime coordinates explicitly.
+            auxmol.set_geom_(mol.atom_coords(unit='Bohr'), unit='Bohr')
         self.nao = mol.nao
         self.naux = auxmol.nao
 
@@ -219,9 +224,14 @@ class DF(lib.StreamObject):
         '''Reset mol and clean up relevant attributes for scanner mode'''
         if mol is not None:
             self.mol = mol
-            self.auxmol = None
-            self.intopt = None
-            self.j_engine = None
+        # auxmol, intopt and j_engine depend on the geometry of self.mol, so
+        # they must be discarded even when reset() is called without a new mol
+        # (e.g. after an in-place mol.set_geom_()). Keeping them rebuilds
+        # _cderi against stale auxiliary-basis coordinates and yields silently
+        # wrong SCF energies (issue #827).
+        self.auxmol = None
+        self.intopt = None
+        self.j_engine = None
         self._cderi = None
         self._cderi_idx = None
         self._cd_j2c = None
--- a/gpu4pyscf/df/tests/test_df_reset.py
+++ b/gpu4pyscf/df/tests/test_df_reset.py
@@ -73,11 +73,35 @@ class KnownValues(unittest.TestCase):
 
         assert abs(e_reused - e_fresh) < 1e-8
 
-    def test_reset_invalidates_geometry_caches(self):
+    def test_reset_preserves_auxmol_reference(self):
+        # Several internal callers (df/grad, hessian, tddft) hold a reference
+        # to with_df.auxmol across reset(), using reset() as a memory-release
+        # idiom. reset() without a geometry change must keep the object (and
+        # its identity) valid.
         mf = dft.RKS(mol.copy(), xc='wb97x').density_fit().to_gpu()
         mf.kernel()
+        aux_ref = mf.with_df.auxmol
+        assert aux_ref is not None
         mf.reset()
-        assert mf.with_df.auxmol is None
+        assert mf.with_df.auxmol is aux_ref
+        # The kept object must still be usable (the CI failure mode read
+        # auxmol.with_range_coulomb right after reset()).
+        with aux_ref.with_range_coulomb(0.3):
+            pass
+
+    def test_reset_repairs_geometry_caches_in_place(self):
+        # After an in-place displacement, reset() re-anchors auxmol to the
+        # runtime coordinates (same object) and drops the stale integral
+        # engines.
+        mf = dft.RKS(mol.copy(), xc='wb97x').density_fit().to_gpu()
+        mf.kernel()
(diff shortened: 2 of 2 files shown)


sunqm (approved):

Isitea:
CI root cause identified and fixed in b0e7c7a.

The 96 `AttributeError: 'NoneType' object has no attribute 'with_range_coulomb'` failures all trace to one semantic conflict: df/grad, hessian and tddft modules use `with_df.reset()` as a **memory-release idiom** while holding a reference to `with_df.auxmol` (e.g. `gpu4pyscf/df/grad/rhf.py`: fetch `auxmol`, call `reset()` to free GPU memory, then use `auxmol`). Discarding `auxmol` unconditionally broke that contract.

Revised approach — **repair in place instead of discarding**. `reset(mol=None)` now:
- geometry unchanged → keeps `auxmol`/`intopt`/`j_engine` exactly as before (only cderi tensors dropped) — fully backward compatible with the memory-release idiom;
- geometry changed in place (`mol.set_geom_`) → re-anchors `auxmol` to the runtime coordinates **preserving object identity**, and drops the geometry-bound integral engines — still fixing #827, without invalidating held references;
- atom-count mismatch (defensive) → discard.

Regression tests updated: one now encodes the held-reference idiom the CI exercised (`auxmol is aux_ref` + `with_range_coulomb` usable right after `reset()`). Re-validated on the 1.7.4 runtime: displaced-reuse error stays at 4.3e-11 Ha and the idiom passes. Branch is also merged up to current master (51a94c9).

Isitea:
Additional real-world validation: we independently hit the identical failure (`'NoneType' object has no attribute 'with_range_coulomb'` during DF geometry optimization) in our production pipeline, which had been running the *original* unconditional-invalidation semantics as a runtime patch on 1.7.4. Switching the patch to the revised repair-in-place semantics from b0e7c7a fixed the pipeline end-to-end (full 5-phase DFT workflow incl. geometry optimization, SMD, and Hessians now running clean on the same code path the CI exercises).

sunqm:
In-place modifying `auxmol` within reset function is dangerous. It can easily lead to unexpected results. This is a misuse of the in-place geometry change. Please see the comments in #827

Isitea:
Thank you for the clarification, and for taking the time to review this.

The reason I initially read this as a bug was that the same sequence happened to produce correct results with CPU PySCF, so the difference in behavior on GPU4PySCF looked like a regression rather than a usage error on my side. Your explanation makes the API contract clear: a mol modified by `set_geom_()` should be treated as a new Mole instance and passed explicitly via `mf.reset(mol)`, rather than expecting dependent objects to pick up in-place changes.

I verified on v1.7.4 that the supported pattern indeed gives the correct result (Δ 4.3e-11 Ha against a freshly built reference), and I have updated our downstream code to follow this convention throughout.

Thanks again for the guidance — much appreciated.

PR #855, opened 2026-08-14
fix and enable fp32 multigrid-v2 collo kernels

Efithor:
for LDA/GGA FP32 instantiations of multigrid-v2 collocation kernels the per-grid-point pair for float the corner value underflows while the recursion start factors overflow, giving 0 * inf = NaN. Fixed via evaluating exponent directly instead of recursion product + these codepaths . Double routes and MGGA drivers are unaffected


gpu4pyscf/lib/multigrid/multigrid_v2/evaluation.cuh
```diff
@@ -299,7 +299,19 @@ __global__ static void evaluate_density_kernel(
           gto_cartesian<KernelType, j_angular>(j_cartesian, x - j_x, y - j_y,
                                                z - j_z);
 
-          const KernelType gaussian = gaussian_x * gaussian_y * gaussian_z;
+          KernelType gaussian = gaussian_x * gaussian_y * gaussian_z;
+          if constexpr (sizeof(KernelType) == 4) {
```
henryw7: I don't think this is a good idea at all, as the algorithm is completely changed, and most of the data preparation part of this kernel should be removed if we simply recompute the gaussian at each grid. I believe simply changing all variables to fp32 is not the right choice for mixed/single precision here.

sunqm:
We don't plan to enable fp32 by default, as its numerical behavior is not guaranteed. You can add a cmake option to enable fp32 for users who want to experiment it in local build.

Efithor:
> We don't plan to enable fp32 by default, as its numerical behavior is not guaranteed. You can add a cmake option to enable fp32 for users who want to experiment it in local build.

@sunqm Done!

sunqm (approved):

PR #862, opened 2026-08-19
Fix a CUDA_ILLEGAL_MEMORY_ACCESS bug for PBC RSDF + large system + multigpu

henryw7:
In https://github.com/pyscf/gpu4pyscf/blob/ff6b09bdc322a5749aefdca7e5822b1cccf19029/gpu4pyscf/pbc/df/rsdf_builder.py#L411 it calls https://github.com/pyscf/gpu4pyscf/blob/ff6b09bdc322a5749aefdca7e5822b1cccf19029/gpu4pyscf/pbc/df/rsdf_builder.py#L346 where https://github.com/pyscf/gpu4pyscf/blob/ff6b09bdc322a5749aefdca7e5822b1cccf19029/gpu4pyscf/pbc/df/int3c2e.py#L387 the `gout_stride` variable always resides on one particular GPU, because of the cache.

I believe the correct solution is to remove all `lru_cache` that could possibly cache a device memory piece, as this is always a trap for multigpu application. And this simple and fast function doesn't require a cache at all.

This bug is only triggered for big enough system, because otherwise only one GPU (device_id = 0 in practice) will get to the `eval_j3c` part, other GPUs will early quit.

### DIFF
--- a/gpu4pyscf/df/int3c2e_bdiv.py
+++ b/gpu4pyscf/df/int3c2e_bdiv.py
@@ -544,7 +544,6 @@ class Int3c2eEnvVars(ctypes.Structure):
         return Int3c2eEnvVars.new(self.natm, self.nbas, atm, bas, env, ao_loc,
                                   self.log_cutoff)
 
-@multi_gpu.lru_cache
 def int3c2e_scheme(*, short_range=False, shm_size=SHM_SIZE, gout_width=None,
                    gout_ndim='ijk', deriv=None, cache_cart_idx=False,
                    angular_inc=None):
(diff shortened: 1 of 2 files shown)


gpu4pyscf/df/int3c2e_bdiv.py
```diff
@@ -544,7 +544,6 @@ def copy(self):
         return Int3c2eEnvVars.new(self.natm, self.nbas, atm, bas, env, ao_loc,
                                   self.log_cutoff)
 
-@lru_cache
 def int3c2e_scheme(*, short_range=False, shm_size=SHM_SIZE, gout_width=None,
```
sunqm, line 547: multi_gpu.lru_cache can be used instead

sunqm (approved):

henryw7:
I got an error
```
  File "/somewhere/gpu4pyscf/gpu4pyscf/pbc/df/int3c2e.py", line 697, in int3c2e_scheme
    return mol_int3c2e_scheme(
           ^^^^^^^^^^^^^^^^^^^
TypeError: lru_cache.<locals>.to_cache() got an unexpected keyword argument 'short_range'
```
using `multi_gpu.lru_cache`?

I believe the correct usage is `multi_gpu.lru_cache(1)`, even if I add default argument to `multi_gpu.lru_cache` function.

PR #869, opened 2026-08-24
Add GPU density-fitted CASCI and diagonal-Hessian CASSCF

wtpeter:
## Summary

This PR adds initial GPU support for density-fitted CASCI and CASSCF.

The main additions are:

- An in-core GPU `direct_spin1` FCI solver with GPU Hamiltonian contraction, Davidson diagonalization, and RDM construction.
- GPU density-fitting AO-to-MO integral transformation through `DF.ao2mo`.
- `DFCASCI` support for restricted HF and KS references.
- A diagonal-Hessian `DFCASSCF` orbital optimizer.
- Timing information for AO2MO, effective one-electron integrals, FCI, RDM construction, orbital derivatives, and orbital rotations.
- Unit tests and a CPU/GPU DF-CASSCF comparison example.

The CASSCF orbital step uses a positive diagonal-Hessian model with a common level shift. The final orbital rotation parameters are limited by `max_stepsize` without increasing the level shift solely to enforce the step limit.

## Current limitations

- Only density-fitted CASCI/CASSCF is supported.
- Only restricted HF/KS references are supported.
- The GPU FCI solver is in-core only and raises `MemoryError` when GPU memory is insufficient.
- CASSCF uses an approximate diagonal Hessian rather than the full augmented-Hessian optimizer.
- Canonicalization, natural orbitals, frozen orbitals, symmetry-adapted FCI, and state-averaged CASSCF are not implemented.

This implementation is intended as a lightweight GPU CASSCF fallback rather than a replacement for the more robust second-order CPU optimizer.

### DIFF
--- a/gpu4pyscf/fci/direct_spin1.py
+++ b/gpu4pyscf/fci/direct_spin1.py
@@ -1,4 +1,4 @@
-# Copyright 2021-2024 The PySCF Developers. All Rights Reserved.
+# Copyright 2026 The PySCF Developers. All Rights Reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -14,7 +14,14 @@
 
 import numpy as np
 import cupy as cp
+
+from pyscf import lib
+from pyscf.fci import cistring
 from pyscf.fci import direct_spin1
+
+from gpu4pyscf.lib import logger
+from gpu4pyscf.lib.cupy_helper import get_avail_mem
+
 
 TILE = 32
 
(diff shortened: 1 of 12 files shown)


gpu4pyscf/fci/direct_spin1.py
```diff
@@ -128,65 +169,413 @@
...
+    if not 0 < nroots <= vector_size:
+        raise ValueError(f'Invalid nroots {nroots} for vector size {vector_size}')
+
+    vector_bytes = vector_size * x0.dtype.itemsize
+    max_space = min(max_space + (nroots - 1) * 4, vector_size)
+    work_vectors = max(5 * nroots, len(x0))
+    required_memory = ((2 * max_space + work_vectors) * vector_bytes +
+                       max_space**2 * x0.dtype.itemsize)
+    available_memory = get_avail_mem()
+    memory_limit = min(max_memory * 1e6, available_memory)
```
sunqm, line 349: `max_memory` in pyscf controls CPU memory. Since GPU memory is assumed to be dedicated to a single process, we can ignore `max_memory` when estimating GPU memory usage.

gpu4pyscf/fci/direct_spin1.py
```diff
@@ -128,65 +169,413 @@
...
+    try:
+        xs = cp.empty((max_space, vector_size), dtype=x0.dtype)
+        ax = cp.empty_like(xs)
+        heff = cp.empty((max_space, max_space), dtype=x0.dtype)
+    except cp.cuda.memory.OutOfMemoryError as err:
+        raise MemoryError(
+            f'Failed to allocate GPU Davidson subspace of size {max_space}') \
+            from err
+
+    converged = cp.zeros(nroots, dtype=bool)
```
sunqm, line 368: converged can be put on host memory, as a numpy array

gpu4pyscf/mcscf/casci.py
```diff
@@ -0,0 +1,152 @@
...
+        energy_core += float(cupy.einsum('ij,ji', core_dm, hcore).real.get())
+        energy_core += float(cupy.einsum('ij,ji', core_dm, corevhf).real.get()) * .5
+
+    h1eff = reduce(cupy.dot, (mo_cas.conj().T, hcore+corevhf, mo_cas))
+    out = h1eff.get(), float(energy_core)
+    if return_corevhf:
+        out += (corevhf,)
+    timing = getattr(casci, 'timing', None)
+    if isinstance(timing, dict):
+        timing['h1e_wall'] = (timing.get('h1e_wall', 0.) +
```
sunqm, line 62: wall time is inaccurate in the GPU code.
If you just need to print cost of individual step, it's recommended to use `gpu4pyscf.lib.logger.timer` functions. If you want to collect the timing data, you need to use the cupy Event object, see  also `gpu4pyscf.scf.jk._TimingCollector`

sunqm (approved):

gpu4pyscf/fci/direct_spin1.py
wtpeter: Convergence flags are now stored and evaluated as a NumPy array on the host.

PR #885, opened 2026-08-31
Fix OOM df-tdrhf gradients module (fix #881)

### LINKED ISSUE #881: TD gradient OOMs with 34 GB free: `_jk_energies_by_dm_factors` commits more than `get_avail_mem` reports
**Version:** `gpu4pyscf-cuda12x` 1.8.1 (`df/grad/tdrhf.py` byte-identical in 1.8.0), CuPy 14.1.1,
CUDA 12.x, cuTENSOR 2.7.0 active. One A100-SXM4-40GB.

## Symptom

A TD-RKS gradient on a 98-atom molecule (nao=1813, def2-TZVP/PBE0, `def2-universal-jkfit`,
`.density_fit()`, TDA singlet root 1) dies with `cudaErrorMemoryAllocation` while **34.40 GB of
device memory is free** and the SCF+TDA working set is under 4 GB.

Instrumented run, sampling `cupy.cuda.runtime.memGetInfo()` alongside every `get_avail_mem` call:

```
before gradient        device free 34.99 GB | pool used 0.09 reserved 3.87
get_avail_mem call 1:  reports 34.40 GB | driver free 34.40 GB   <- agrees exactly
at failure             device free  4.25 GB | pool used 1.04 reserved 1.31
```

So `get_avail_mem` is not misreporting: the routine is told 34.40 GB, consumes ~30 GB, and then
fails asking for more.

```
File "gpu4pyscf/grad/tdrks.py", line 276, in grad_elec
  ejk = td_grad.jk_energies_per_atom(...)
File "gpu4pyscf/df/grad/tdrhf.py", line 366, in _jk_energies_per_atom
  out.append(_jk_energies_by_dm_factors(
File "gpu4pyscf/df/grad/tdrhf.py", line 544, in _jk_energies_by_dm_factors
  compressed = cp.zeros((nao_pair, naux_in_batch))
cupy_backends.cuda.api.runtime.CUDARuntimeError: cudaErrorMemoryAllocation: out of memory
```

## Cause

In `_jk_energies_by_dm_factors`, `batch_size` is chosen at line 407 so that the `buf` allocation
takes half the available budget:

```python
mem_avail  = mem_free - 2*naux*np.dot(dm1_noccs, dm2_noccs)*8 - 2*n_dm*nao**2*8   # 406
batch_size = int(mem_avail*.5/(n_dm*nao_pair*8))                                   # 407
```

`buf` is then allocated at that size (line 534/538) and stays live for the whole batch loop:

```python
buf = cp.empty((n_dm*nao_pair*batch_size))   # 538
```

But inside the loop, `compressed` is allocated per batch at essentially the same shape, and this
allocation is not reflected in the `batch_size` calculation:

```python
for kbatch, lk in enumerate(uniq_l_ctr_aux[:,0]):
    naux_in_batch = nf[lk] * l_ctr_aux_counts[kbatch]
    compressed = cp.zeros((nao_pair, naux_in_batch))   # 544  <- not budgeted for
```

`batch_size` is derived as if `buf` were the only consumer of that half of `mem_avail`, so with
`compressed` live at the same time the pair alone approaches the whole budget — before the three
`blksize`-sized arrays (`j3c_full`, `buf1`, `buf2`, lines 421/529/530, sized from the *same*
`mem_avail` at line 416) are counted.

Note the failure mode is worse on a *larger* card: the over-commit is a fraction of whatever
`get_avail_mem` reports, so more free memory means a proportionally larger over-commit.

## Workaround

Halving the figure the routine is given is enough, on the same system that OOMs uncapped:

```python
import gpu4pyscf.df.grad.tdrhf as tdrhf_df
_orig = tdrhf_df.get_avail_mem
tdrhf_df.get_avail_mem = lambda *a, **k: int(_orig(*a, **k) * 0.5)
```

Same geometry, same build, one input changed:

| | uncapped | scaled by 0.5 |
|---|---|---|
| `get_avail_mem` reports | 34.40 GB | 17.20 GB |
| driver free at that moment | 34.40 GB | 34.40 GB |
| outcome | OOM at line 544, 4.25 GB free | **completes** |
| device free afterwards | — | 37.39 GB |

Note what this implies: the device still holds all 34.40 GB in the capped run. Halving only the
*reported* figure fixes it, so the demand is a multiple of what the routine is told rather than
anything near the hardware limit. Peak pool reservation on the successful run was 1.46 GB, and
results are unchanged since they are summed over blocks.

## Suggested fix

Account for `compressed` where `batch_size` is derived — it is live alongside `buf` for the whole
loop, so the `.5` at line 407 should be closer to `.25`, or `compressed` should reuse `buf`'s
storage as the non-`sum_results` branch effectively does.

sunqm:


michalkrompiec:
This rebinds `buf1`, which already holds a live scratch buffer, so `compressed` ends up aliased with the views taken from it inside the loop. On a small DF TD gradient it silently returns a wrong answer.

### The aliasing

Before the contraction loop, two scratch buffers are bound:

```python
buf1 = cp.empty((blksize, nao, nao))
buf2 = cp.empty((blksize, nao, nao))
```

The `sum_results` branch then rebinds the first:

```python
buf, buf1 = cp.empty((2, nao_pair*batch_size))
```

`buf1` still has two readers further down, neither guarded by `sum_results`:

```python
dm_tensor1 = ndarray((nao,nao,dk), buffer=buf1)
tmp        = ndarray((dm2_noccs[i],nao,dk), buffer=buf1)
```

After the rebind these alias the new `compressed`, which is also a view of `buf1` and starts at the same offset. `compressed` is zeroed once per `kbatch` and accumulated across the inner loop, but

```python
dm_tensor1[:] = dm_tensor.transpose(1,0,2)
dm_tensor1[:] += dm_tensor
```

overwrites that memory on every iteration, and the `compressed[:,k0:k1] += cp.take(...)` that follows reads a source overlapping its own destination. What reaches `sum_ejk_int3c2e_ip1` is not the contracted tensor.

### Measured

CH2O / def2-SVP / PBE0, `RKS.density_fit()`, TDA, gradient of state 1, on an A100 with gpu4pyscf 1.8.1. Three runs, each in a fresh process, differing only in this one file:

| variant | max abs diff vs. stock (Ha/Bohr) |
| --- | --- |
| stock 1.8.1 | reference |
| this PR | **20.57** |
| this PR, second buffer renamed off `buf1` | 1.7e-10 |

Stock gives `dE/dz` = +0.1167 (C) and -0.1167 (O); this PR gives -14.55 and +20.45, and the translational sum rule is off by 20.5 where the other two satisfy it to 1e-13. Renaming the rebound buffer restores the stock result exactly, so the batching change itself is fine — only the reuse of the name is the problem.

Note that this PR **does not raise**. The out-of-bounds writes described below land inside the memory pool without faulting, so the result is a silently wrong gradient.

### The rebound buffer can also be too small

`buf1` was `blksize*nao*nao` doubles and becomes `nao_pair*batch_size`. `dm_tensor1` needs `nao*nao*dk` with `dk <= blksize`, and `nao_pair ~ nao^2/2`, so the view fits only when `batch_size >= 2*blksize`.

When memory is plentiful both `batch_size` and `blksize` clamp to `naux`, and the buffer is about half what is needed. From the formulas in this function:

| case | batch_size | blksize | `buf1` has | `dm_tensor1` needs | ratio |
| --- | --- | --- | --- | --- | --- |
| H2O / def2-SVP, 40 GB free | 84 | 84 | 25,200 | 48,384 | 0.52 |
| CH2O / def2-SVP, 40 GB free | 132 | 132 | 97,812 | 190,608 | 0.51 |
| 98 atoms / def2-TZVP, 34 GB | 646 | 288 | 1.06e9 | 9.47e8 | 1.12 |

`lib.cupy_helper.ndarray` builds views from a raw `memptr` with no bounds check, and `buf1` is the second row of `cp.empty((2, ...))`, so the excess writes run past the end of that allocation. Small systems are the ones that hit this; at production size the views fit and only the aliasing bites.

### One thing to watch in the sizing

Rebinding `buf1` drops the original `(blksize, nao, nao)` array and returns `0.225*mem_avail` to the pool, so the branch's apparent footprint falls. Counting what is live in the contraction phase, as a fraction of `mem_avail`, with `n_dm = 3` (what `grad/tdrhf.py` passes on the TD-gradient path):

| | total |
| --- | --- |
| current master | 0.78 |
| this PR | 0.73 |
| this PR with `buf1` left intact | 0.95 |

So the reduction comes from the buffer being released rather than from the batching change. Worth noting too that `buf1` and `buf2` together are `0.45*mem_avail` and are not subtracted from it anywhere.

michalkrompiec:
Suggested fix:
```
--- a/gpu4pyscf/df/grad/tdrhf.py
+++ b/gpu4pyscf/df/grad/tdrhf.py
@@ -550,7 +550,7 @@
     if sum_results:
         kern = libvhf_rys.sum_ejk_int3c2e_ip1
         ejk = cp.zeros((mol.natm, 3))
-        buf, buf1 = cp.empty((2, nao_pair*batch_size))
+        buf, buf3 = cp.empty((2, nao_pair*batch_size))
     else:
         kern = libvhf_rys.ejk_int3c2e_ip1
         ejk = cp.zeros((n_dm, mol.natm, 3))
@@ -560,7 +560,7 @@
         naux_in_batch = nf[lk] * l_ctr_aux_counts[kbatch]
         aux_ao_offset = aux_loc[ksh_offsets_cpu[kbatch]]
         if sum_results:
-            compressed = ndarray((nao_pair, naux_in_batch), buffer=buf1)
+            compressed = ndarray((nao_pair, naux_in_batch), buffer=buf3)
             compressed.fill(0.)
         else:
             compressed = ndarray((n_dm, nao_pair, naux_in_batch), buffer=buf)
```

PR #886, opened 2026-09-01
[SYCL] Infrastructure to support SYCL backend

abagusetty:
Thanks to @sunqm and @alvaro for this effort


gpu4pyscf/lib/cupy_helper/add_sparse.cu
```diff
@@ -38,13 +50,23 @@ extern "C" {
 __host__
 int add_sparse(cudaStream_t stream, double *a, double *b, int *indices, int n, int m, int count){
     int ntile = (m + THREADS - 1) / THREADS;
+#ifdef USE_SYCL
+    sycl::range<2> threads(THREADS, THREADS);
```
sunqm, line 54: Instead of using `#if USE_SYCL` in every cu file, is it possible to introduce a few helper functions/macros to simplify the kernel launch? For example, something like `auto grid = make_grid(...)`, `auto block = make_blocks(...)`, and `launch_kernel(kernel_name, grid, block, arguments, ...)`.

sunqm:
I've created a `sycl-backend` branch. Let’s first target this PR to that branch and run the github pipelines there. Once we resolved most of the issues, we can merge it into the main branch.

gpu4pyscf/lib/cupy_helper/add_sparse.cu
```diff
@@ -22,8 +22,20 @@
 __global__
 void _add_sparse(double *a, double *b, int *indices, int n, int m, int count)
 {
-	int row = blockIdx.x * BLOCK_DIM + threadIdx.x;
-    int col = blockIdx.y * BLOCK_DIM + threadIdx.y;
+#ifdef USE_SYCL
+    auto item = syclex::this_work_item::get_nd_item<2>();
+    int blockIdx_x = item.get_group(1);
```
sunqm, line 27: I think `blockIdx_x` and `blockIdx_y` ... could be defined as macros. We could then use these names directly in the code. This could  reduce these duplicated `#ifdef` blocks.

abagusetty:
> I've created a `sycl-backend` branch. Let’s first target this PR to that branch and run the github pipelines there. Once we resolved most of the issues, we can merge it into the main branch.

Sure sounds good. Will close this here

PR #887, opened 2026-09-02
Becke grid XC nuclear derivative, including grid response

henryw7:
- [x] KRKS
- [x] KUKS

Additionally, in KUKS CDIIS, fixed a bug that when basis set linear dependency is detected and nao != nmo, the contraction calls in the following code receives `out` variables with wrong shape:
https://github.com/pyscf/gpu4pyscf/blob/70a48ab6e39b12ab9dd86c19989368350f40175a/gpu4pyscf/scf/diis.py#L98:L106


sunqm (approved):

gpu4pyscf/pbc/grad/rhf.py
```diff
@@ -149,7 +149,7 @@ def energy_ee(self, dm):
             j_factor = 0
         elif xc.upper() != 'HF':
             from gpu4pyscf.pbc.grad.krks import get_vxc
-            de += get_vxc(ni, mf.cell, mf.grids, xc, dm[None], np.zeros((1, 3))) * 2
+            de += get_vxc(ni, mf.cell, mf.grids, xc, dm[None], np.zeros((1, 3)))
```
sunqm, line 152: This is a wrong fix. The test `test_lda_grad` in test_pbc_grad_rks.py would fail after disabling `multigrid_numint()`

sunqm (changes requested):

sunqm (approved):

PR #894, opened 2026-09-03
cleanup cp vs np usage, early return with warp-divergence in ft_ao.cu

abagusetty:
1. Mostly involves minor replacing cp vs np usage
2. Fix a warp divergence that luckily works in CUDA, but not generic for HIP/SYCL

### DIFF
--- a/.github/workflows/unittest.yml
+++ b/.github/workflows/unittest.yml
@@ -59,4 +59,4 @@ jobs:
           -v $GITHUB_WORKSPACE:/workspace \
           -v ~/.cache/pip:/root/.cache/pip \
           pyscf/gpu4pyscf-devel:pyscf-2.14 \
-          /bin/bash -c "cd /workspace && pip3 install -r requirements.txt && pip3 install pyscf==2.8 scipy==1.17 && source build.sh && pytest -m 'not slow and not benchmark and not special' --cov=/workspace --durations=50 && rm -rf .pytest_cache"
+          /bin/bash -c "cd /workspace && pip3 install -r requirements.txt && pip3 install MCFun && source build.sh && pytest -m 'not slow and not benchmark and not special' --cov=/workspace --durations=50 && rm -rf .pytest_cache"
--- a/.github/workflows/unittest.yml
+++ b/.github/workflows/unittest.yml
@@ -59,4 +59,4 @@ jobs:
           -v $GITHUB_WORKSPACE:/workspace \
           -v ~/.cache/pip:/root/.cache/pip \
           pyscf/gpu4pyscf-devel:pyscf-2.14 \
-          /bin/bash -c "cd /workspace && pip3 install -r requirements.txt && pip3 install MCFun && source build.sh && pytest -m 'not slow and not benchmark and not special' --cov=/workspace --durations=50 && rm -rf .pytest_cache"
+          /bin/bash -c "cd /workspace && pip3 install -r requirements.txt && pip3 install pyscf==2.8 scipy==1.17 && source build.sh && pytest -m 'not slow and not benchmark and not special' --cov=/workspace --durations=50 && rm -rf .pytest_cache"
--- a/gpu4pyscf/df/tests/test_df_hessian.py
+++ b/gpu4pyscf/df/tests/test_df_hessian.py
@@ -358,7 +358,7 @@ class KnownValues(unittest.TestCase):
         mo_occ = mf.mo_occ
         test_hessian_round2 = hobj.partial_hess_elec(mo_energy, mo_coeff, mo_occ)
 
-        assert np.max(np.abs(test_hessian_round1 - test_hessian_round2)) < 2e-7
+        assert abs(test_hessian_round1 - test_hessian_round2).max().item() < 2e-7
 
 if __name__ == "__main__":
     print("Full Tests for DF Hessian")
--- a/gpu4pyscf/lib/pbc/ft_ao.cu
+++ b/gpu4pyscf/lib/pbc/ft_ao.cu
@@ -49,16 +49,15 @@ void ft_ao_bdiv_kernel(double *out, RysIntEnvVars envs, int nGv, double *Gv)
     int sh_id_in_block = threadIdx.y;
     int Gv_id_in_block = threadIdx.x;
     int sh_id = sh_block_id * nsh_per_block + sh_id_in_block;
-    if (sh_id >= envs.nbas) {
-        return;
-    }
+    int valid = sh_id < envs.nbas;
+    int sh_id_clamped = valid ? sh_id : envs.nbas - 1;
 
     int *atm = envs.atm;
     int *bas = envs.bas;
     double *env = envs.env;
-    int li = bas[sh_id*BAS_SLOTS+ANG_OF];
+    int li = bas[sh_id_clamped*BAS_SLOTS+ANG_OF];
     int nfi = c_nf[li];
-    int iprim = bas[sh_id*BAS_SLOTS+NPRIM_OF];
+    int iprim = bas[sh_id_clamped*BAS_SLOTS+NPRIM_OF];
     int Gv_id = Gv_block_id * NG_PER_BLOCK + Gv_id_in_block;
     double kx = 0;
     double ky = 0;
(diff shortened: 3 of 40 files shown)


gpu4pyscf/df/tests/test_df_hessian.py
```diff
@@ -358,7 +358,7 @@ def test_unstable_j2c(self):
         mo_occ = mf.mo_occ
         test_hessian_round2 = hobj.partial_hess_elec(mo_energy, mo_coeff, mo_occ)
 
-        assert np.max(np.abs(test_hessian_round1 - test_hessian_round2)) < 2e-7
+        assert abs(test_hessian_round1 - test_hessian_round2).max().item() < 2e-7
```
henryw7: I'm ok with this change, I just wonder why this one is different from others, as I'm using np.max(np.abs()) everywhere, and only this one got changed.

.github/workflows/unittest.yml
```diff
@@ -59,4 +59,4 @@ jobs:
           -v $GITHUB_WORKSPACE:/workspace \
           -v ~/.cache/pip:/root/.cache/pip \
           pyscf/gpu4pyscf-devel:pyscf-2.14 \
-          /bin/bash -c "cd /workspace && pip3 install -r requirements.txt && pip3 install pyscf==2.8 scipy==1.17 && source build.sh && pytest -m 'not slow and not benchmark and not special' --cov=/workspace --durations=50 && rm -rf .pytest_cache"
+          /bin/bash -c "cd /workspace && pip3 install -r requirements.txt && pip3 install MCFun && source build.sh && pytest -m 'not slow and not benchmark and not special' --cov=/workspace --durations=50 && rm -rf .pytest_cache"
```
sunqm, line 62: Please don't change these settings. They were intentionally configured this way, to ensure code still working for old versions.

gpu4pyscf/pbc/dft/multigrid_v2.py
```diff
@@ -1398,7 +1398,7 @@ def convert_xc_on_g_mesh_to_fock_gradient(
 def get_nuc(ni, kpts=None):
     if ni.sorted_gaussian_pairs is None:
         ni.build()
-    is_single_kpt = kpts is not None and kpts.ndim == 1
+    is_single_kpt = kpts is None or kpts.ndim == 1
```
sunqm, line 1401: Do not modify this

gpu4pyscf/pbc/dft/multigrid_v2.py
```diff
@@ -1417,7 +1417,7 @@ def get_pp(ni, kpts=None):
     """Get the periodic pseudopotential nuc-el AO matrix, with G=0 removed."""
     if ni.sorted_gaussian_pairs is None:
         ni.build()
-    is_single_kpt = kpts is not None and kpts.ndim == 1
+    is_single_kpt = kpts is None or kpts.ndim == 1
```
sunqm, line 1420: Do not modify this

gpu4pyscf/pbc/dft/multigrid_v3.py
```diff
@@ -1535,7 +1535,7 @@ def get_rho(ni, dm_kpts, kpts=None):
 
 def get_nuc(ni, kpts=None):
     cell = ni.cell
-    is_single_kpt = kpts is not None and kpts.ndim == 1
+    is_single_kpt = kpts is None or kpts.ndim == 1
```
sunqm, line 1538: Do not modify this

gpu4pyscf/pbc/dft/multigrid_v3.py
```diff
@@ -1561,7 +1561,7 @@ def get_pp(ni, kpts=None):
     log = logger.new_logger(cell)
     t0 = log.init_timer()
 
-    is_single_kpt = kpts is not None and kpts.ndim == 1
+    is_single_kpt = kpts is None or kpts.ndim == 1
```
sunqm, line 1564: Do not modify this

gpu4pyscf/lib/pbc/ft_ao.cu
```diff
@@ -72,6 +71,7 @@ void ft_ao_bdiv_kernel(double *out, RysIntEnvVars envs, int nGv, double *Gv)
 
     int gx_len = (AUXL+1) * FT_AO_THREADS;
     __shared__ double g[(AUXL+1)*FT_AO_THREADS * 6];
+    __shared__ int block_iprim[FT_AO_THREADS/NG_PER_BLOCK];
```
sunqm, line 74: Preferable to redefine NG_PER_BLOCK, making it align to warp size. Introducing this block_iprim looks unnecessary.

.github/workflows/unittest.yml
abagusetty: Done, revert to original

gpu4pyscf/pbc/dft/multigrid_v2.py
abagusetty: Done, revert to original

gpu4pyscf/pbc/dft/multigrid_v2.py
abagusetty: Done, revert to original

gpu4pyscf/pbc/dft/multigrid_v3.py
abagusetty: Done, revert to original

gpu4pyscf/pbc/dft/multigrid_v3.py
abagusetty: Done, revert to original

gpu4pyscf/lib/pbc/ft_ao.cu
abagusetty: addressed this, This made it a much cleaner solution for both CUDA and SYCL

gpu4pyscf/lib/pbc/ft_ao.cu
```diff
@@ -30,8 +30,11 @@
 #endif
 #define WARPS           8
 #define THREADS         256
-#define NG_PER_BLOCK    WARP_SIZE
 #define FT_AO_THREADS   (WARP_SIZE*4)
+// One shell per block (nsh_per_block == 1): every thread in the block then
+// sees the same shell's iprim, so the primitive loop's __syncthreads() trip
+// count is uniform without needing a per-block max-iprim workaround.
+#define NG_PER_BLOCK    FT_AO_THREADS
```
sunqm, line 37: I think it's better to define`#define WARP_SIZE 64` for SYCL backend instead, leaving the cuda warp configurations unchanged.

abagusetty: May be the comments were confusing. SYCL (on Intel GPUs support 32 warpsize) and it already uses the same CUDA `WARP_SIZE` settings. There is no change here. The use of `WARP_SIZE 64` may be needed in the future to support HIP backends. Not the scope yet. 

Before we had this: 
```c++
#define WARP_SIZE       32
#define NG_PER_BLOCK    WARP_SIZE          // = 32
#define FT_AO_THREADS   (WARP_SIZE*4)      // = 128

__global__ static
void ft_ao_bdiv_kernel(...)
{
    ...
    int nsh_per_block = FT_AO_THREADS / NG_PER_BLOCK;   // = 4
    int sh_id = sh_block_id * nsh_per_block + sh_id_in_block;
    if (sh_id >= envs.nbas) {
        return;
    }
    ...
    int iprim = bas[sh_id*BAS_SLOTS+NPRIM_OF];   // <-- differs per shell
    ...
    for (int ip = 0; ip < iprim; ++ip) {
        __syncthreads();          // <-- barrier, trip count = iprim
        ...
    }
```

CUDA: It is tolerated in CUDA for all the threads (32 of them in a warp) to not participate in `__syncthreads()`. It is a UB but works. There is no hang even if only threads in a warp do `__syncthreads()`

SYCL: is strict about all the threads in a warp to reach `__syncthreads()`. Any divergence leads to hangs because only some SYCL threads in a warp do `__syncthreads()` and other dont. HIP also needs this uniform behavior across the thread-group in the future. 

So the change was to:
```c++
#define FT_AO_THREADS   (WARP_SIZE*4)      // = 128
#define NG_PER_BLOCK    FT_AO_THREADS      // = 128, was WARP_SIZE

__global__ static
void ft_ao_bdiv_kernel(...)
{
    int nsh_per_block = FT_AO_THREADS / NG_PER_BLOCK;   // = 1, was 4
    ...
    int iprim = bas[sh_id_clamped*BAS_SLOTS+NPRIM_OF];   // same shell for whole group
    ...
    for (int ip = 0; ip < iprim; ++ip) {
        __syncthreads();          // trip count now uniform across the group
        ...
    }
```


There is no performance or functionality changes here for CUDA

abagusetty: @sunqm Shelving this for now, reverting to original to keep it simple

PR #905, opened 2026-09-09
multigrid_v3._partition_ke_for_fft - prevent shell pairs whose Ecut estimate was clamped to ke_cutoff from falling outside FFT buckets by clamping the estimate to the largest f32 not above ke_cutoff

Efithor:
multigrid_v3._partition_ke_for_fft - prevent shell pairs whose Ecut estimate was clamped to ke_cutoff from falling outside FFT buckets by clamping the estimate to the largest f32 not above ke_cutoff


sunqm:
multigrid_v3 implementation has been updated. This fix is not needed.

PR #907, opened 2026-09-11
Fix GPU OOM in PBC GDF gradient computation

puzhichen:
Fix GPU out-of-memory errors in PBC GDF gradient:
- Add proper memory budgeting for both SR (_get_j3c_block_sizes) and LR (_get_lr_block_size) parts, accounting for all working buffers under complex128 and raising clear errors when memory is insufficient.
- Explicitly release intermediate tensors (None) for earlier CuPy memory reuse.
- Add a gamma-poin path that delegates to the real-valued RHF implementation when k-points and density matrices are real.


gpu4pyscf/pbc/df/grad/kuhf.py
```diff
@@ -77,6 +79,30 @@ def _get_ejk_derivatives(int3c2e_opt, dm, kpts=None, hermi=0, j_factor=1., k_fac
     t0 = log.init_timer()
 
     dm_factor_l, dm_factor_r = factorize_dm(dm, hermi)
+    if (hermi == 1 and len(kpts) == 1 and is_zero(kpts) and bvk_ncells == 1 and
+        (j_factor == 0 or not omega) and
+        dm_factor_l.dtype == np.float64 and
+        (dm_factor_r is None or dm_factor_r.dtype == np.float64)):
+        # Exchange has no cross-spin terms. Process each spin through the
```
sunqm: Just call the uhf._get_ejk_derivatives function

sunqm (approved):

